PR #27, opened 2016-08-29
Feature: `ppc_resid_scatter`

jgabry:
@paul-buerkner Ok just added the scatterplot of y vs resids on the `feature-ppc_resid_scatter` branch. I think it's working so whenever you have a chance let me know what you think and I'll merge this PR once we’re satisfied.


paul-buerkner:
Looks good. Can we also have a `ppc_resid_scatter_avg` plot? Or do you think it does not make sense?

jgabry:
On Monday, August 29, 2016, Paul-Christian Bürkner <notifications@github.com>
wrote:

> Looks good. Can we also have a ppc_resid_scatter_avg plot? Or do you
> think it does not make sense?
>

I can add that.

>

paul-buerkner:
That would be great!

PR #50, opened 2016-11-19
use png for html vignette figures

tjmahr:
The figures are not showing up in the CRAN vignettes. 

If you need both pdf and png output, try using a vector with `c("png", "pdf")` in the `dev` option.

### DIFF
--- a/vignettes/MCMC-diagnostics.Rmd
+++ b/vignettes/MCMC-diagnostics.Rmd
@@ -18,7 +18,7 @@ library("ggplot2")
 library("gridExtra")
 library("rstan")
 knitr::opts_chunk$set(
-  dev = "pdf",
+  dev = "png",
   fig.align = "center",
   fig.width = 4,
   fig.height = 3,
--- a/vignettes/MCMC.Rmd
+++ b/vignettes/MCMC.Rmd
@@ -14,7 +14,7 @@ vignette: >
 ```{r, settings, include=FALSE}
 library(bayesplot)
 knitr::opts_chunk$set(
-  dev = "pdf",
+  dev = "png",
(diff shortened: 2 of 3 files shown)


jgabry:
Thanks, glad you spotted that! The figures look fine if opening the
vignette locally after installing from CRAN but they definitely aren't
rendering in the version hosted on CRAN's website. I was using pdf for the
images because it's a pain in the ass to get properly sized high quality
png figures using the html_vignette format with knitr and rmarkdown. The
default gives poor quality images (see for example the images in the
rstanarm vignettes on the CRAN website), but if you increase the dpi then
you have to manually adjust all figure sizes, and then if you want to
change the dpi you have to do that again every time. That was getting super
annoying. Any tips on how I can get same quality images as the ones using
pdf and with the same figure sizes?

tjmahr:
I follow the tips here: http://r4ds.had.co.nz/graphics-for-communication.html#figure-sizing 

Yesterday, I [posted about plotting fits](https://tjmahr.github.io/visualizing-uncertainty-rstanarm/) from RStanARM models 😎 , and I used the following knitr options:

```r
  opts_chunk$set(
    fig.asp = 0.618,
    fig.width = 6,
    dpi = 300,
    fig.align = "center",
    out.width = "80%")
```

To customize specific plots, then, you'd manipulate either `out.width` or `fig.asp` to adjust the size or dimensions.

Also, I tried it a bit with vignettes here. The frustrating thing is how long it can take to re-render the vignette after tweaking one of these options.

jgabry:
Thanks for the tips. I'll play around with the settings a bit based on those recommendations and see how it goes.

PR #57, opened 2016-12-12
Introduce mcmc_recover_intervals

### LINKED ISSUE #56: plots to facilitate comparing mcmc estimates to "true" parameter values
After fitting a model to simulated data (using fixed parameter values) plot a comparison of the MCMC estimates  and the "true" parameter values to check that they are (approximately) recovered.

jgabry:
For plotting a comparison of MCMC estimates and "true" parameter values (values used to simulate the data) to check that they are (approximately) recovered.

closes #56


jgabry:
@ericnovik check out this new `mcmc_recover_intervals` function

PR #62, opened 2017-01-02
Feature/ppc rootogram

paul-buerkner:
This PR implements rootograms for count data models.


R/ppc-rootograms.R
```diff
@@ -45,28 +47,31 @@ ppc_rootogram <- function(y, yrep, style = c("standing", "suspended"),
...
+  if (style == "hanging")
+    ypos <- tyexp - ypos
 
-  data <- data.frame(x, ty, tyexp, tyquantile)
+  data <- data.frame(xpos, ypos, ty, tyexp, tyquantile)
   graph <- ggplot(data) +
-    aes_(ymin = ~ tylower, ymax = ~ tyupper) +
-    geom_col(
-      aes_(x = ~ x, y = ~ ty, fill = "Observed"),
+    aes_(ymin = ~ tylower, ymax = ~ tyupper, height = ~ ty) +
```
jgabry, line 108: @paul-buerkner Is there a reason that `aes_` is added to the ggplot object rather than specified as an argument to the `ggplot` function? I didn't even know that was possible. Seems to work I guess.

paul-buerkner: No particular reason. You can change it if you want.

jgabry: Aside from `x`, which of the aesthetics are used by both `geom_tile` and `geom_smooth`?

paul-buerkner: Non, I believe. The thing is that `ymin` and `ymax` are used by `geom_smooth` but can currently not be specified within it (this will be fixed in the next release of **ggplot2**). Same with `height` for `geom_tile`.

jgabry:
Ah ok, let's leave it as is for now then.

jgabry:
@paul-buerkner thanks for adding this!

PR #69, opened 2017-01-30
ggplot + grid implementation of rstan's pairs plot

### LINKED ISSUE #67: ggplot + grid implementation of rstan's pairs plot
This is basically done on the `mcmc_pairs` branch but still needs to be cleaned up and tested.

jgabry:
This PR introduces `mcmc_pairs` and `mcmc_hex`. Closes #67.

### Example usage
From `help("mcmc_pairs")`:

#### mcmc_hex

```r
x <- example_mcmc_draws(params = 6)
dimnames(x)
(p <- mcmc_hex(x, pars = c("sigma", "alpha"), transformations = list(sigma = "log")))
p + plot_bg(fill = "gray95")
p + plot_bg(fill = "gray95") + panel_bg(fill = "gray70")
```

#### mcmc_pairs
```r
color_scheme_set("purple")
# default of condition=NULL implies splitting chains between upper and lower panels
mcmc_pairs(x, pars = "alpha", regex_pars = "beta\\[[1,4]\\]",
           off_diag_args = list(size = 1, alpha = 0.5))

# change to density plots instead of histograms and hex plots instead of
# scatterplots
mcmc_pairs(x, pars = "alpha", regex_pars = "beta\\[[1,4]\\]",
           diag_fun = "dens", off_diag_fun = "hex")

# plot chain 1 separately from chains 2, 3, and 4
mcmc_pairs(x, pars = "alpha", regex_pars = "beta\\[[1,4]\\]",
           diag_fun = "dens", off_diag_fun = "hex",
           condition = pairs_condition(chains = list(1, 2:4)))
```

and with NUTS diagnostic info (divergences, max_treedepth, etc.)
```r
# pairs plot with NUTS diagnostic info overlaid (example using rstanarm package)
library(rstanarm)

# for demonstration purposes, intentionally fit a model that
# will (almost certainly) have some divergences
fit <- stan_glm(
  mpg ~ ., data = mtcars,
  iter = 1000,
  # this combo of prior and adapt_delta should lead to some divergences
  prior = hs(),
  adapt_delta = 0.9
)
post <- as.array(fit)
np <- nuts_params(fit)

# split the draws according to above/below median accept_stat__ and
# show approximate location of divergences (red points)
mcmc_pairs(
  post,
  pars = c("wt", "cyl", "sigma"),
  off_diag_args = list(size = 1, alpha = 0.5),
  condition = pairs_condition(nuts = "accept_stat__"),
  np = np
)

# using median log-posterior as 'condition', hex instead of scatter for
# off-diagonal plots, and also indicating where max treedepth hit
color_scheme_set("darkgray")
mcmc_pairs(
  post,
  pars = c("wt", "cyl", "sigma"),
  off_diag_fun = "hex",
  condition = pairs_condition(nuts = "lp__"),
  lp = log_posterior(fit),
  np = np,
  np_style = pairs_style_np(div_color = "firebrick", td_color = "dodgerblue", div_size = 1, td_size = 2),
  # for demonstration purposes, set max_treedepth to a value that will
  # result in at least a few warnings points (colored with np_style$color[2], i.e. "dodgerblue" in this case)
  max_treedepth = with(np, max(Value[Parameter == "treedepth__"]) - 1)
)
```


jgabry:
@paul-buerkner Any thoughts on this pairs plot implementation?

bgoodri:
Haven't gotten to it yet

jgabry:
@paul-buerkner Hey Paul, any thoughts on this pairs plot implementation before I proceed forward with it?

paul-buerkner:
I am taking a look at it right now.

jgabry:
Cool. Thanks Paul!

paul-buerkner:
I really love the new pairs plot. It is thoroughly documented and all options seem to work well. Three minor comments:

- The input style for arguments `condition` and `np_style` is maybe not ideal in my opinion as it does not help users with the appropriate structure. Do you think it makes sense to add some helper functions helping users setting up these arguments?
- Is plotting a line on the y-axis necessary for the diagonal plots given that no y-axis labels are used?
- Is it possible to align the x-axis of the diagonal and the off diagonal plots per column? I assume this is too complicated, but I wanted to ask anyway.

jgabry:
@paul-buerkner Thanks for the comments! A few responses: 

> - The input style for arguments `condition` and `np_style` is maybe not ideal in my opinion as it does not help users with the appropriate structure. Do you think it makes sense to add some helper functions helping users setting up these arguments?

Yeah I agree it's not ideal. What kind of helper functions are you envisioning? Alternatively, any ideas for alternative ways to specify those arguments? 

> - Argument name `transformations` is rather long. Could we shorten it somehow?

It _is_ long! And I kind regret using `transformations` (instead of maybe `trans` or `transform`), but it's the same for all of the other `mcmc_*` functions that accept the argument. So I don't want to change it just for `mcmc_pairs`. The good news is that it comes before the `...` so using a partial name is allowed (e.g., `trans = list(...)` should work). I'm open to changing the argument name in all of the functions to something shorter, but I'm not sure it's worth deprecating it and introducing a shorter form. What do you think?

> - Doc of `transformations`: the length of the supplied list does not need to be the length of the parameters (as suggested by the doc). Instead it can have any length but the names of the list elements should match (some of) the parameters. Also there is a typo ("transfomred") in the doc.

Aside from the "transfomred" typo, doesn't it already say that it should have the same length as the number of parameters to be transformed (not the total number of parameters)? I guess technically it could have more elements than number of parameters to be transformed (so long as the additional elements don't match parameter names) but I'm not sure if that's work mentioning. 

> - Is plotting a line on the y-axis necessary for the diagonal plots given that no y-axis labels are used?

Not necessary. I think removing it is a good idea. It should make the grid of plots look a bit cleaner.

> - Is it possible to align the x-axis of the diagonal and the off diagonal plots per column? I assume this is too complicated, but I wanted to ask anyway.

Yeah I thought about this. I think it could probably be done but I'm not sure how to do it. It's probably something that would require quite a bit of fiddling with the stuff in the `grid` package. Maybe something to implement in the future if someone (you or I or someone else) has the time to figure it out?

paul-buerkner:
> Yeah I agree it's not ideal. What kind of helper functions are you envisioning? Alternatively, any ideas for alternative ways to specify those arguments?

For `np_style` we could have a function with three arguments (color, shape, size), which would be more structured than a named list of variable length and include default arguments as well. Also, one could include argument checks already in this helper function.

For `conditions` one may add a function with arguments `chains_lower`, `chains_upper` taking the input options related to chains, `prob_lower` for the proportion of samples for the lower part and argument `param` to specify possible NUTS parameters (argument names not ideal yet). All arguments could default to `NULL`. This would ease the interpretations of the input as it would be explicit through argument names and not implicit through input type.

>  I'm open to changing the argument name in all of the functions to something shorter, but I'm not sure it's worth deprecating it and introducing a shorter form. What do you think?

I missed that other functions use this argument as well. I don't think it is worth deprecating as we have partial matching.

> Aside from the "transfomred" typo, doesn't it already say that it should have the same length as the number of parameters to be transformed (not the total number of parameters).

You are right, the doc is correct. I misread that one sentence.

> Maybe something to implement in the future if someone (you or I or someone else) has the time to figure it out?

I agree. It might be possible somehow, but it will get rather complicated as the axis position depends on the digits on the y-axis.

jgabry:
Cool, thanks. I do like the helper function idea. I'll try to add them
soon.

jgabry:
@paul-buerkner What do you think of the new `pairs_style_np` and `pairs_condition` helper functions?

paul-buerkner:
Looks good! A few minor comments:

- The output of `pairs_conditions` is unstructured and, as far as I understand, `mcmc_pairs` still has to make sense of it based on the input type (numeric, logical, list, etc.). I am not sure if that makes the code simpler, but one may want to use a list output instead (similar to `pairs_style_np`), possibly with a special element naming the condition type.
- There is double "the" in the doc of `chains, draws, nuts`

jgabry:
Ok thanks, yeah I just added a "type" attribute to the object returned by `pairs_condition` which makes the branching logic easier to follow when actually handling the condition.

PR #74, opened 2017-02-28
ppc_violin_grouped: plot y as violin, too

silberzwiebel:
I plotted a reaction time analysis using ppc_violin_grouped and got this:

pointdata

However, I personally find this plot more informative as it shows the density of the empirical data also:

violindata

I created a pull request to deliver the solution to my problem instead of just opening an issue and `complaining`. Possibly, however, there are reasons out of my knowledge why the data is plotted as `geom_point` and not `geom_violin`. 
I'm also not sure about any possible conventions regarding the plotting style of data. I found the current style easy to see for my example.

### DIFF
--- a/R/ppc-distributions.R
+++ b/R/ppc-distributions.R
@@ -88,6 +88,10 @@
 #' ppc_violin_grouped(y, yrep, group, size = 1.5)
 #' \donttest{
 #' ppc_violin_grouped(y, yrep, group, alpha = 0)
+#' 
+#' # examples to show non-default options of ppc_violin_grouped() for y-drawing 
+#' ppc_violin_grouped(y, yrep, group, alpha = 0, ydraw = "point", size = 2, yalpha = 0.5, jitter = 0.2)
+#' ppc_violin_grouped(y, yrep, group, alpha = 0, ydraw = "both", size = 2, yalpha = 0.5, jitter = 0.2)
 #' }
 #'
 NULL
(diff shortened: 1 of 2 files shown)


jgabry:
Thanks! I like it. Now I'm wondering if it's worth offering both options (y
as violin, y as points) or just going with the violin. Or perhaps both at
the same time. When there is a lot of data the points probably aren't that
useful, but when y doesn't contain too many observations I do think seeing
the actual points can be helpful.

avehtari:
If you decide to have an option for showing the points, it would be useful to add some jitter in x direction, so that it's possible to see if there are several overlapping y values. This will work also for small number of observations.

jgabry:
Good idea. In the version with points I'll definitely jitter them a bit
horizontally.

silberzwiebel:
Cool. I'll implement an option that let's the user specify if he/she wants to see y as violin, jittered points, or both, in the next week or so.
This would also add some more substance to this pull request that for now only changed 2 lines ;)

Regarding the documentation: Where would I need to put the description of the additional options so that it lands in the documentation of `violin_grouped`? Until now, I didn't code inside an R package.

jgabry:
Great, thanks! 

The documentation for a new argument to `ppc_violin_grouped` can go right after [these lines](https://github.com/stan-dev/bayesplot/blob/master/R/ppc-distributions.R#L349-L351) in the `ppc-distributions.R` file (right above the definition of the `ppc_violin_grouped` function). You can add it right below the documentation for the `probs` argument that's already there. The format is like this:

```r
#' @param argname One or several sentences describing what the argument is used for. 
```
where `argname` is whatever the new argument name is. 

Once you add that you can run 
```r
devtools::document()
devtools::load_all()
?ppc_violin_grouped
```
and you should see the updated documentation.

silberzwiebel:
Just added the jittered points and an option to toggle between plotting `y` as violin, jittered points or both. I've also added some more arguments to manipulate appearance of `y` (defaults based on best-looking for some of mine real-world models) and two examples. Tested everything via `brms::pp_check()`. Not sure, if I followed all potential code guideline, so feel free to edit where necessary.

jgabry:
I have some small changes I want to make but this looks good! I'm going to merge it now and then I'll make the changes, which is probably easier than me making commits to your branch

PR #86, opened 2017-04-20
Convert parameter_names.matrix to parameter_names.default

helske:
Small change, previously I was able to use `mcmc` objects from `coda` with `bayesplot`, but due to the change of `parameter_names` into S3 method, this is no longer possible, as there are only methods for objects of class `matrix` and `array`. As `mcmc` object is basically vector or matrix, the matrix method should be applicable to `mcmc` objects as well (in matrix case). And there are perhaps other similar cases. So I changed the `parameter_names.matrix` to default method, which shouldn't break anything but makes the method and an all the function which use it more robust.


jgabry:
Thanks for the PR! That issue hadn't occurred to me. Your change looks fine.

PR #112, opened 2017-09-08
Divergences for mcmc_scatter

jgabry:
This PR adds an `np` and `np_style` arguments to `mcmc_scatter` so that divergences can be highlighted in the plot. It also adds the `scatter_style_np` function to use when specifying `np_style`. 

(It's a little bit annoying to have a bunch of functions (e.g. `scatter_style_np`, `pairs_style_np`, `trace_style_np`, `parcoord_style_np`) that are used to specify the `np_style` argument depending on the plotting function. It would be nice just to have one function for that, but I tried playing around with that and it's even more annoying because it makes it more confusing to have different defaults for the shape, style, alpha, arguments depending on the plotting function. With different functions the default values can be made more explicit.) 

#### Example usage

```r

library(rstan)
fit <- stan_demo("eight_schools")
posterior <- as.matrix(fit)

color_scheme_set("darkgray")
mcmc_scatter(
  posterior, 
  pars = c("tau", "theta[1]"), 
  np = nuts_params(fit), 
  np_style = scatter_style_np(div_color = "green", div_alpha = 0.8)
)
```

scatter-with-divs


jgabry:
(Running R CMD CHECK now, will report back with the results)

jgabry:
Passes R CMD CHECK

jgabry:
Cool. I just added a quick example of using `mcmc_scatter` with the `np` argument in the `visual-mcmc-diagnostics.Rmd` vignette. In the previous PR for `mcmc_parcoord` I also added an example of that to the same vignette. Eventually it will probably make sense to break up the current vignettes into smaller ones if/when they get too long.

Issue #97, opened 2017-07-26
Expose some functions that return data used for plotting


jgabry:
Suggested by @tjmahr. Currently the ggplot objects returned by bayesplot can be modified using many of the ggplot2 functions but some things are hard or impossible to change once the object has been created. Providing access to some of the functions that prepare the data would give users much more flexibility.

jgabry:
I've received a few questions about why we would expose the `_data` functions if the data is already stored in the ggplot object, so I'm adding some comments here to clarify a bit (and so I can point people here when they ask that question). 

Here are two of the reasons for adding the `_data` functions:

1. While it's true that the data is stored inside the ggplot object, only in simple cases is it convenient to retrieve the data from the ggplot object. 

For example: 

```r
library("ggplot2")
g <- ggplot(mtcars, aes(wt, mpg)) + geom_point()
```

In this case `g$data` will be identical to the `mtcars` data frame. However, now consider the following example: 

```r
library("dplyr")
d1 <- mtcars %>% filter(wt > 3)
d2 <- mtcars %>% filter(wt <= 3)
g <- ggplot(d1, aes(wt, mpg)) + 
  geom_point(color = "purple") + 
  geom_point(data = d2, color = "green")
```

Now `g$data` is only identical to `d1` and to get `d2` we'd need to do something like `g$layers[[2]]$data`. In this simple example we could have just used `mtcars` and set `aes(color = wt > 3)` but for more complicated plots it can be useful to provide the user access to the data used for plotting in a convenient form rather than require the user to go on a scavenger hunt inside the ggplot object. 

2. In addition to being exposed to the user, the `_data` functions will be used inside the __bayesplot__ plotting functions. Currently everything that the `_data` functions will do exists inside the __bayesplot__ plotting functions, so creating the `_data` functions is just a way of separating the data prep from the plotting, which will let us test these two parts separately. Even though there will technically be more functions to test, it doesn't actually add to the maintenance burden (if anything it probably reduces it).

PR #115, opened 2017-09-20
refactor mcmc_intervals() and mcmc_areas()

### LINKED ISSUE #103: refactor mcmc_intervals/mcmc_areas
These are based on very old code and the same functionality could be provided using less code (and simpler code). Some improvements could be made too, for example using a discrete scale for the vertical axis.

tjmahr:
Closes #103.

Advances #97  (add `_data()` functions).

These plots now use a discrete y-axis. Previously, we used a continuous y-axis, but relabeled the integer labels with parameter names. This caused some unexpected behavior. 

We also use ggridges to draw the density curves in order to keep a discrete y-axis. One possibly unexpected implementation detail is that the point estimate in mcmc_areas() is no longer a vertical line segment, but a very narrow interval drawn by ggridges as well.

The main visual differences are:

* these plots no longer have a "forehead" for an extra step on the y-axis. (This could in principle be restored)
* that the density curves are _not_ normalized to have the same height. 

Previously, a flat, thin density curve would be stretched vertically and look like a wall.

image

Now it looks more like a flat density curve and the spikes in the other parameters look like spikes.

image

(Of course, neither of these plots is any good, because of the variables being plotted.)

## Visual diffs

These are the plots in the examples. 

_Current_
01-current
_Update_
01-updated
***
_Current_
02-current
_Update_
02-updated
***
_Current_
03-current
_Update_
03-updated
***
_Current_
04-current
_Update_
04-updated
***
_Current_
05-current
_Update_
05-updated
***
_Current_
06-current
_Update_
06-updated
***
_Current_
07-current
_Update_
07-updated
***
_Current_
08-current
_Update_
08-updated
***
_Current_
09-current
_Update_
09-updated
***
_Current_
10-current
_Update_
10-updated
***
_Current_
11-current
_Update_
11-updated


tjmahr:
Hmm, we need a bit of forehead back or we need to slightly tweak the vertical scaling. I'm noticing some of the areas get truncated.

image

tjmahr:
Okay, slightly expanded y-axis for mcmc_areas.

image

jgabry:
I haven't had a chance to go through the code thoroughly yet, but at first
glance this all looks great!

There seems to be a segfault with Travis on R-devel but that was happening
before this PR, so we'll deal with that separately. My hunch is that it's
related to RStan but I haven't looked into it yet.

Anyway, I'll take a more thorough look at the PR later today or tomorrow
and either suggest some changes or merge it.

DESCRIPTION
```diff
@@ -25,7 +25,9 @@ Imports:
     reshape2,
     stats,
     utils,
-    rlang
+    rlang,
+    tidyr,
+    ggridges
```
jgabry: I'm always hesitant to add more dependencies, especially when only used in one or two places. Do you think __tidyr__ will be useful for other stuff going forward too? If so then I think it's fine to add the dependency now, but if not then is it possible to implement this without tidyr?

R/helpers-gg.R
```diff
@@ -1,5 +1,53 @@
...
+#' xs <- -2:2
+#' maybe_vertical_line <- if (0 > min(xs) && 0 < max(xs)) {
+#'   vline_0(color = "gray90", size = 0.5)
+#' } else {
+#'   geom_ignore()
+#' }
+geom_ignore <- function(...) {
+  geom_blank(mapping = NULL, data = NULL,
+             show.legend = FALSE, inherit.aes = FALSE)
+}
```
jgabry, line 21: The `geom_ignore` wrapper is a nice idea.

R/mcmc-intervals.R
```diff
@@ -146,274 +218,345 @@ mcmc_areas <- function(x,
...
-      ),
-      size = 4,
-      shape = 21
-    )
-    if (!color_by_rhat) {
-      point_args$color <- get_color("dh")
-      point_args$fill <- get_color("l")
-    }
+# Don't import `filter`: otherwise, you get a warning when using
+# `devtools::load_all(".")` because stats also has a `filter` function
```
jgabry, line 410: Yeah that's annoying. It'd be nice to import `filter` from __dplyr__ since we use it a lot but I think it would require importing only what we use from __stats__ as opposed to  just having `@import stats`. We could potentially do that.

DESCRIPTION
tjmahr: I think I figured out a way to do it without tidyr.

tjmahr: Got it

tjmahr:
`compute_column_density()` is kind of intense with its nonstandard evaluation, but it supports arbitrary groups. E.g., for ridgelines... :sunglasses: 

```r
library(ggplot2)
library(ggridges)
devtools::load_all(".")

chains <- example_mcmc_draws() %>%
  prepare_mcmc_array() %>%
  melt_mcmc() %>%
  compute_column_density(c(Parameter, Chain), Value) %>%
  dplyr::filter(Parameter != "alpha", Parameter != "sigma")

ggplot(chains) +
  aes(x = x, y = Parameter) +
  geom_density_ridges(aes(height = density, color = factor(Chain)),
                      stat = "identity", fill = NA)
```

image

DESCRIPTION
jgabry: Thanks. I'm totally fine with using tidyr now if you think you'll probably want to use it a lot going forward. If not then I think the way you found of doing it without tidyr looks fine. Sorry if it was a pain to figure out.

PR #148, opened 2018-04-30
Pass 'breaks' to geom_histogram

jgabry:
This PR just adds a `breaks` argument to histograms that can be used as an alternative to `binwidth`. Turns out to be useful sometimes.


jgabry:
@tjmahr I requested your review on this to keep with best practice, but I know you're short on time these days and this is a super simple PR, so if you want me to just merge it that's fine. Otherwise review it whenever you have the chance, but no rush.

tjmahr:
Yup, super simple PR.

Issue #120, opened 2017-09-27
infer annotation aesthetics from global theme


tjmahr:
I was thinking today about how `mcmc_intervals()` draws a vertical line if the x's contain 0.

```r
library(bayesplot)
library(ggplot2)
x <- example_mcmc_draws()
mcmc_intervals(x)
```
![image](https://user-images.githubusercontent.com/1890315/30932774-1c0ee2e6-a38e-11e7-9b58-a86b98b2ca80.png)

That vertical line doesn't play well with other themes.

```r
theme_set(theme_grey())
mcmc_intervals(x)
```

![image](https://user-images.githubusercontent.com/1890315/30932784-23ad7a08-a38e-11e7-93d7-5f470181de98.png)

That line color is hard-coded... Specifically in this code.

```r 
# faint vertical line at zero if zero is within x_lim
layer_vertical_line <- if (0 > x_lim[1] && 0 < x_lim[2]) {
  vline_0(color = "gray90", size = 0.5)
} else {
  geom_ignore()
}
```

Gridlines are made invisible by changing their size, so can recover their color.

But in principle, we could just borrow theme's current grid-line color and fatten it.

```r
layer_vertical_line <- if (0 > x_lim[1] && 0 < x_lim[2]) {
  # this would change to `theme_bayesplot_get()` if go that route in 
  # issue 117
  t <- theme_get()
  # `x %||% y` returns y when x is NULL and x otherwise
  color <- t$panel.grid.major$colour %||% "grey92"
  minor_size <- t$panel.grid.minor$size %||% .125
  major_size <- t$panel.grid.major$size %||% (minor_size * 2)
  size <- major_size * 2
  vline_0(color = color, size = size)
} else {
  geom_ignore()
}
```

When there are no gridlines, it would default to .5.

```r
theme_set(theme_default())
mcmc_intervals(x)
```

![image](https://user-images.githubusercontent.com/1890315/30934388-16b83ca2-a393-11e7-992e-09e2e06ff9f4.png)

Otherwise it goes twice the width of the major gridlines.

```r
theme_set(theme_grey())
mcmc_intervals(x)
```

![image](https://user-images.githubusercontent.com/1890315/30934409-25792ea4-a393-11e7-90ab-9621ecfafc53.png)

And generalize to any other gridlined theme.

```r
theme_set(ggthemes::theme_solarized())
mcmc_intervals(x)
```

[...]

PR #149, opened 2018-05-02
Add bayesplot_theme_*() functions and remove startup theme_set()

### LINKED ISSUE #117: setting global theme: costs vs. benefits
It took me a few minutes to discover why all the figures in our paper changed. Answer: one of the co-authors added some code that used bayesplot. It took another few minutes of googling to figure out how to fix that. Not a huge deal, but it's nice when loading and using a package that there are as few unintended side effects as possible. On the other hand, there are benefits if you are only using bayesplot and/or like the theme. For me, the (small) costs outweigh the benefits.

Thoughts on the cost/benefit tradeoff of setting the global ggplot theme?

malcolmbarrett:
Closes #117

This PR adds `bayesplot_theme_set()`, `bayesplot_theme_get()`, `bayesplot_theme_update()`, and `bayesplot_theme_replace()`, which apply only to `bayesplot`s. The user can change the themes applied to `bayesplot` using these functions, which are now the default for all plotting functions. `bayesplot` no longer changes the default `ggplot2` theme, instead using this internal approach, and `bayesplot_theme_*()` functions do not affect other `ggplot`s.

The `ggplot2` equivalents supersede these. Loading `ggplot2` will not affect `bayesplot` themes, but explicitly calling e.g. `theme_set(theme_minimal())` will change all `ggplot`s, including `bayesplot`s, to `theme_minimal()`. I believe that this is the functionality most people would want, but this is done with a single line of code checking if the `ggplot2` theme is set to anything other than `theme_grey()` (the default) and can easily be removed.


jgabry:
@malcolmbarrett Thanks so much for working on this! We'll go through and review it as soon as possible.

tjmahr:
In the final case, the bayesplot is still using the dark theme.

```r
library(bayesplot)
library(ggplot2)

# plot using the default theme automatically 
x <- example_mcmc_draws()
mcmc_hist(x)

# change theme for all ggplots
theme_set(theme_dark())
mcmc_hist(x)

# change back to the default
bayesplot_theme_set(theme_default())
mcmc_hist(x)
```

tjmahr:
In `mcmc_intervals()` and others, we do some light theming on the plot. For example, we move the legend to the top and bold the y-axis font and tick labels. Applying the current theme at the end undoes those theme changes.

```
  ggplot(data) +
    layer_vertical_line +
    layer_outer +
    layer_inner +
    layer_point +
    scale_color +
    scale_fill +
    scale_y_discrete(limits = unique(rev(data$parameter))) +
    xlim(x_lim) +
    legend_move(ifelse(color_by_rhat, "top", "none")) +
    yaxis_text(face = "bold") +
    yaxis_title(FALSE) +
    yaxis_ticks(size = 1) +
    xaxis_title(FALSE) +
    bayesplot_theme_get()
```

We should apply the bayesplot theme earlier so that the theming works. We should also limit/cut back on this kind of by-plot theming and document any custom-theming (that can be its own issue). (I think there are good uses for by-plot theming, like in the ridgeline plots where it's good have horizontal lines and to align the axis labels above the tick marks.)

malcolmbarrett:
@tjmahr that example isn't in the docs, right? In that case, setting `theme_set(theme_grey())` would reset it, but that's a little out of the way. I did think it was possible that there would be a small subset of people that would want different themes set for regular ggplots and bayesplots (that seems strange to me, but maybe I'm wrong), in which case it may make sense to have something like `override_theme_set(TRUE)` or something with `options` to always use `bayesplot_theme_get()` regardless of the set ggplot theme. That could be embedded in bayesplot_theme_set, e.g. `bayesplot_theme_get(theme_default(), override = TRUE)`... but it all seems like such a specific use case to me. 

Regarding your second comment: I shuffled the theme calls around so they are now in the correct order.

tjmahr:
>  I did think it was possible that there would be a small subset of people that would want different themes set for regular ggplots and bayesplots (that seems strange to me, but maybe I'm wrong), in which case it may make sense to have something like override_theme_set(TRUE) or something with options to always use bayesplot_theme_get() regardless of the set ggplot theme.

I'm thinking more in terms of the code working as expected... `bayesplot_theme_set()` should *always* update the bayesplot theme, even if `theme_set()` has already specified a global ggplot2 theme.

malcolmbarrett:
When you put it that way, I think you're absolutely right. Try this. It should work the way you describe. The only case I can come up with where it's not totally in order is something like this:

```r 
#  change all themes
theme_set(theme_dark())
mcmc_hist(x)

# change back to the default
bayesplot_theme_set(theme_default())
mcmc_hist(x)

# try to change all themes again to the same theme as above 
theme_set(theme_dark())
mcmc_hist(x) # still theme_default()

# but setting a new theme works:
theme_set(theme_minimal())
mcmc_hist(x) # theme_minimal()
```

This is because it now saves the existing ggplot2 `theme_get()` and compares it the next time `bayesplot_get_theme()` is run, but there's no way to tell that the second `theme_dark()` is different. Still, that's a bit of a silly situation, and if you really want it back to `theme_dark()`, you can do it with `bayesplot_theme_set()`. Thoughts?

jgabry:
@malcolmbarrett Thanks again for this PR! I haven't had a chance to run the code yet (will do soon), but some quick comments:

> I did think it was possible that there would be a small subset of people that would want different themes set for regular ggplots and bayesplots (that seems strange to me, but maybe I'm wrong)

I actually don't think it's that strange but that may be because I've seen every single bayesplot plot  many many times. Some of them aren't as easily readable if certain kinds of themes are set. So I actually think there can be a good reason for maintaining a separate theme for bayesplot, although most plots will be fine regardless of the theme. 

@tjmahr I also definitely agree with both of these other items you mentioned: 

> We should also limit/cut back on this kind of by-plot theming and document any custom-theming (that can be its own issue).

> bayesplot_theme_set() should _always_ update the bayesplot theme, even if theme_set() has already specified a global ggplot2 theme

jgabry:
Oh, and I know it would be slightly different behavior than `ggplot2::theme_set()`, but I think it would be nice if `bayesplot_theme_set()` defaulted to setting `bayesplot::theme_default()` instead of throwing an error if no argument is specified. Having to do `bayesplot_theme_set(theme_default())` in order to reset the default seems unnecessary. Thoughts?

malcolmbarrett:
Ok, I'm convinced! Try out the solution I committed above and see what you think.

> Having to do bayesplot_theme_set(theme_default()) in order to reset the default seems unnecessary.

Makes sense to me! Just added.

jgabry:
@malcolmbarrett Thanks for adding the default. 

@tjmahr Any else you think should be changed before we merge this?

tjmahr:
@jgabry Let me look at this Wednesday. I'll run the visual unit tests too. We should start to think about the ideas in #120 for our next steps.

jgabry:
@tjmahr sounds good (and no rush if you’re super busy at the moment)

R/bayesplot-ggplot-themes.R
```diff
@@ -77,3 +72,105 @@ theme_default <-
...
+#' # change back to the default
+#' bayesplot_theme_set() # same as bayesplot_theme_set(theme_default())
+#' mcmc_areas(x, regex_pars = "beta")
+#'
+#' # change theme for all ggplots
+#' theme_set(theme_dark())
+#' mcmc_dens_overlay(x)
+#'
+bayesplot_theme_get <- function() {
+  if (!identical(.bayesplot_theme_env$gg_current, ggplot2::theme_get())) {
```
tjmahr, line 132: This is cool

R/mcmc-diagnostics-nuts.R
```diff
@@ -262,14 +264,16 @@ mcmc_nuts_divergence <- function(x, lp, chain = NULL, ...) {
   violin_lp <- ggplot(violin_lp_data, aes_(x = ~ Value, y = ~ lp)) +
     geom_violin(fill = get_color("l"), color = get_color("lh")) +
     ylab("lp__") +
-    xaxis_title(FALSE)
+    xaxis_title(FALSE) +
+      bayesplot_theme_get()
```
tjmahr, line 269: I had thought that `bayesplot_theme_get()` would override the change by `xaxis_title()` because `xaxis_title()` can theme xaxis titles. But `xaxis_title(FALSE)` just calls `labs()` so these are all fine.

jgabry:
Just double checking, are we ready to finally merge this?

tjmahr:
I think so!

Issue #144, opened 2018-04-22
Extend vignettes


jgabry:
The vignettes haven’t been updated
in a while. They need some updating to reflect the content added to the package since they were last worked on. 

Also, @martinmodrak any interest in adding some content from or similar to  your blog post 

http://www.martinmodrak.cz/2018/02/19/taming-divergences-in-stan-models/

to give more substance to the diagnostics vignette?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

martinmodrak:
Could probably do that, but is this the right place? If the dev team believes the contents of the blog post is correct and illustrative enough to be part of official materials, shouldn't that go to Stan manual or RStan vignettes? 

Right now, the blog does not show how to use diagnostic visualization (which I hope it sometimes will, but you know how those plans go). If I had something to say directly about visaulization, that would certainly make sense here.

tjmahr:
I'll eventually write a vignette to cover bayesplot for ggplot2 users... That would demonstrate the data tidying features and some use cases for the `_data()` functions, like combining draws from multiple models.

jgabry:
Ok no rush or obligation! Just wanted to let you know that I liked the post and that we definitely encourage contributions 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

martinmodrak:
OK, so I did some changes to the diagnostic vignette. My main goal was to present the function in the order I would run them when diagnosing a modelling issue. In that view, it might make sense to move the "Diagnostics for the No-U-Turn Sampler" to the top, before "General MCMC diagnostics", at least I personally use the NUTS diagnostics much more often than Rhat/n_eff (I look at those in the sampling output). But I am not certain about this.

PR #153, opened 2018-05-17
Restructuring and clarifying the diagnostic vignette

martinmodrak:
As discussed in #144, I made some changes to the diagnostic vignette to (IMHO) be more close to a modeller's workflow.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -5,7 +5,8 @@ Version: 1.5.0.9000
 Date: 2018-03-29
 Authors@R: c(person("Jonah", "Gabry", role = c("aut", "cre"), email = "jsg2201@columbia.edu"),
              person("Tristan", "Mahr", role = "aut"),
-             person("Paul-Christian", "Buerkner", role = "ctb"))
+             person("Paul-Christian", "Bürkner", role = "ctb"),
+             person("Martin", "Modrák", role = "ctb"))
 Maintainer: Jonah Gabry <jsg2201@columbia.edu>
 Description: Plotting functions for posterior analysis, model checking,
     and MCMC diagnostics. The package is designed not only to provide convenient
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,10 @@
 
 (GitHub issue/PR numbers in parentheses)
 
+* The [Visual MCMC Diagnostics vignette](http://mc-stan.org/bayesplot/articles/visual-mcmc-diagnostics.html) 
+has been reorganized and has a lot of useful new content  
+thanks to Martin Modrák. (#144, #153)
+
 * The [LOO predictive
 checks](http://mc-stan.org/bayesplot/reference/PPC-loo.html) now require 
 **loo** version `>= 2.0.0`. (#139)
--- a/R/mcmc-diagnostics.R
+++ b/R/mcmc-diagnostics.R
@@ -629,8 +629,8 @@ new_neff_ratio <- function(x) {
 
 validate_neff_ratio <- function(x) {
   stopifnot(is.numeric(x), !is.list(x), !is.array(x))
-  if (any(x < 0 | x > 1, na.rm = TRUE)) {
-    stop("All neff ratios must be between 0 and 1.", call. = FALSE)
+  if (any(x < 0, na.rm = TRUE)) {
+    stop("All neff ratios must be positive.", call. = FALSE)
   }
   x
 }
--- a/R/mcmc-diagnostics.R
+++ b/R/mcmc-diagnostics.R
@@ -216,7 +216,7 @@ mcmc_neff <- function(ratio, ..., size = NULL) {
   check_ignored_arguments(...)
   data <- mcmc_neff_data(ratio)
 
-  max_ratio <- max(ratio)
+  max_ratio <- max(ratio, na.rm = TRUE)
   if(max_ratio < 1.25) {
     additional_breaks <- numeric(0)
   } else if(max_ratio < 1.5) {
--- a/R/mcmc-diagnostics.R
+++ b/R/mcmc-diagnostics.R
@@ -216,6 +216,17 @@ mcmc_neff <- function(ratio, ..., size = NULL) {
   check_ignored_arguments(...)
   data <- mcmc_neff_data(ratio)
 
+  max_ratio <- max(ratio)
+  if(max_ratio < 1.25) {
+    additional_breaks <- numeric(0)
+  } else if(max_ratio < 1.5) {
+    additional_breaks <- 1.25
+    additional_labels <- "1.25"
+  } else {
+    additional_breaks <- seq(1.5, max_ratio, by = 0.5)
+  }
+  breaks <- c(0, 0.1, 0.25, 0.5, 0.75, 1, additional_breaks)
+
   ggplot(
     data,
     mapping = aes_(
@@ -236,9 +247,9 @@ mcmc_neff <- function(ratio, ..., size = NULL) {
     scale_fill_diagnostic("neff") +
     scale_color_diagnostic("neff") +
     scale_x_continuous(
-      breaks = c(0, 0.1, 0.25, 0.5, 0.75, 1),
-      labels = c("0", "0.1", "0.25", "0.5", "0.75", "1"),
-      limits = c(0, 1.05),
+      breaks = breaks,
+      labels = as.character(breaks), #as.character truncates trailing zeroes, while ggplot default does not
+      limits = c(0, max(1, max_ratio) + 0.05),
       expand = c(0, 0)) +
     yaxis_text(FALSE) +
     yaxis_title(FALSE) +
--- a/tests/testthat/test-mcmc-diagnostics.R
+++ b/tests/testthat/test-mcmc-diagnostics.R
@@ -32,7 +32,7 @@ test_that("rhat and neff plot functions throw correct errors & warnings", {
   expect_error(mcmc_rhat(c(-1, 1, 1)), "must be positive")
 
   # need ratios between 0 and 1
-  expect_error(mcmc_neff(c(-1, 0.5, 0.7)), "must be between 0 and 1")
+  expect_error(mcmc_neff(c(-1, 0.5, 0.7)), "must be positive")
 
   # drop NAs and warn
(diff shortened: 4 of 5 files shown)


jgabry:
@martinmodrak Actually I was thinking about your comment in #144 about starting with the NUTS diagnostics before stuff like Rhat, and you're right that it would be closer to actual workflow. Do you want to take a stab at that before I merge this?

martinmodrak:
I'll have a go at it. I also notice some minor typos etc. However, I need to keep the seed (or add seed argument) as the traceplot zoom example may break with another seed. Note that `set.seed` has effect on Stan, because the `seed` argument of `sampling` defaults to `sample.int(.Machine$integer.max, 1)`

jgabry:
I forgot that was the default we have for the seed argument. It won’t
reproduce across different systems but should reproduce on your own system.
Anyway, feel free to add back in the set.seed call.

martinmodrak:
I didn' realize `set.seed` won't reproduce across systems. So I moved to setting the `seed` argument of `sampling`. Also there was a bug as the `n_eff` extractor forces `n_eff / n_samples` to be < 1, which is no longer true in recent Stan (see https://github.com/stan-dev/stan/issues/2450), so I fixed that as well (I didn't bothe rcreating a separate branch for this fix, hop its OK). That's probably all from me, unless you have some other comments.

jgabry:
(Accidentally closed this, so reopening)

martinmodrak:
I initially added the seeds primarily for the traceplot zoom example - I want to focus on the area where divergences arise. While working on the vignette, I also noticed multiple uncommon situations where the vignette ceases to be illustrative:

* Very few or no divergences for the centered parametrization
* Small `Rhat` for all parameters with the short chain example
* One or two divergences in the non-centered parametrization

While I understand that all the conditions for reproducibility have to hold to *guarantee* reproducibility, I would guess that in practice, setting seed would be enough for the results to reproduce well unless the Stan core changes (the RNG in Stan is hardcoded so should not be system dependent). Also we could probably setup some checks and make the vignette fail in those weird cases with something like "Low Rhat in this example, change seed" to make sure the vignette stays relevant.

jgabry:
That all makes sense. I do think R’s set.seed works the same across systems, so you could go back to using that if you want. That won’t guarantee the same Stan results but your proposed solution of forcing it to fail in some cases is a good one I think.

martinmodrak:
Checks added.

jgabry:
@martinmodrak I think this is ready for me to merge, but I had to change one of the seeds to avoid getting a divergence in the NCP model. Does the new seed I changed it to work ok on your end?

avehtari:
Couple comments

1)
> Also there was a bug as the n_eff extractor forces n_eff / n_samples to be < 1, which is no longer true in recent Stan (see stan-dev/stan#2450), so I fixed that as well

To test how the plots look like with high n_eff / n_samples you can use this model
```
parameters {
  vector[16] x;
}
model {
  x ~ normal(0, 1);
}
```
which has n_eff / n_samples > 2.

2)
8 schools NCP has often divergences with the default adapt_delta. Instead of trying to choose random seed (that would work on all platforms), I would use adapt_delta=0.95 (or higher) already for the first fit to make the vignette to produce more consistent results.

avehtari:
> Those functions require more information than simply the posterior samples,
> ...
> That is, even though the samples for tau returned from Stan
> ...
> to obtain samples from the funnel distribution
> ...
> and no samples in the tail of the distribution
> ...
> the ratio of the average variance of samples within each chain to the variance of the pooled samples across chains
> ...
> (it means samples are not independent)

- samples -> draws (following the recommendation by Andrew: a sample is a collection of draws)

> when tau, the hierarchical standard deviation, goes to zero

- goes near zero?

> Then theta is computed deterministically from the parameters eta,mu and tau

- add space after "eta,"

> This is the motivation for the Gelman and Rubin (1992) potential scale reduction statistic

- split-$\hat{R}$ which is used has changed substantially since this paper, and it would be better to use term split-$\hat{R}$ and cite Gelman et al (2013) and Stan manual.

> The effective sample size

- cite Gelman et al (2013) and Stan manual for the algorithm used to compute n_eff

> Autocorrelation is bad (it means samples are not independent) and you want it to drop quickly to zero with increasing lag.

- samples -> draws
- negative autocorrelation is not bad, and non-independent draws can be more efficient than independent

avehtari:
Clarification: I don't suggest adding the normal model to the vignette. Just check that plots look good with n_eff / n > 2. I guess I should have had that comment somewhere else, but since that n_eff fix was also in this pull request...

jgabry:
Thanks Aki. In particular, I think this is a good idea: 

> 8 schools NCP has often divergences with the default adapt_delta. Instead of trying to choose random seed (that would work on all platforms), I would use adapt_delta=0.95 (or higher) already for the first fit to make the vignette to produce more consistent results.

@martinmodrak Are you ok with that?

martinmodrak:
Thanks Aki for the suggestions. I changed all references to "samples" as individual rows in the Stan output to "draws". And I added `adapt_delta` for the initial fits.

The scary thing is that I obviously was wrong about reproducibility and Stan produces mutually different results on my PC, Jonah's PC and the CI server :-O

I am a bit uncertain, how to discuss negative autocorrelation - in particular, I don't really understand how `n_eff > n` actually works - how can you get better than independent samples from the posterior? Does it mean that for any/some function(s) `f`, `f(draws)` approaches the `f(true_target)` faster than `f(independent_draws_from_true_target)`? Or is `n_eff` actually a measure of something a bit different than the number of independent samples from the posterior? Also too much negative autocorrelation would also be bad, wouldn't it? How to check that there is "healthy" negative autocorrelation? Do you have a formulation in mind that would be clear? Otherwise I'll need some time to understand the literature on that to not give misleading interpretations.

I modified the note about Rhat (and it no longer is a citation from the manual). However, please, do not refer to sources by author and year only - for people outside the niche (like me), it is often hard to determine which paper/book is meant. I guess you meant BDA 3?

I have adjusted the `mcmc_neff` plot to not truncate `n_eff > 1`. Sorry for mangling the `n_eff` fix with this pull request.

avehtari:
> I am a bit uncertain, how to discuss negative autocorrelation - in particular, I don't really understand how n_eff > n actually works - how can you get better than independent samples from the posterior? 

See a simple example in 
http://andrewgelman.com/2018/01/18/measuring-speed-stan-incorrectly-faster-thought-cases-due-antithetical-sampling/

>  Does it mean that for any/some function(s) f, f(draws) approaches the f(true_target) faster than f(independent_draws_from_true_target)

Yes. There are examples here http://discourse.mc-stan.org/t/n-eff-bda3-vs-stan/2608/19

> Also too much negative autocorrelation would also be bad, wouldn't it?

No. But you can't get correlation -1 in lag 1, except for binary case

> Do you have a formulation in mind that would be clear? 

Unfortunately I know too much and formulations are likely to be clear for me, but not necessarily for others.

> Otherwise I'll need some time to understand the literature on that to not give misleading interpretations.

I think it's good idea if check those couple links, and then continue discussion to arrive a formulation which would be clear also for wider audience. 

> However, please, do not refer to sources by author and year only - for people outside the niche (like me), it is often hard to determine which paper/book is meant. I guess you meant BDA 3?

Sorry, I guessed it would be clear as it was already listed in the references :)

Also, great that you are interested in improving the documentation!

martinmodrak:
OK, let me give it a try, I would add a footnote to the explanation of `n_eff`. The statement about other functions computed from draws is a guess - does it make sense?:

`n_eff > N` means that the mean value of the parameter computed from Stan draws approaches the true mean faster then independent samples from the true posterior (the estimate from Stan has smaller variance). This is possible when the draws are anticorrelated - draws above the mean tend to be well matched with draws below the mean. Other functions computed from draws (quantiles, posterior intervals, tail probabilities) may not necessarily approach the true posterior faster. Google "antithetic sampling" or visit http://discourse.mc-stan.org/t/n-eff-bda3-vs-stan/2608/19 for some further explanation.

And for the autocorrelation plot:

Positive autocorrelation is bad (it means the chain tends to stay in the same area between iterations) and you want it to drop quickly to zero with increasing lag. Negative autocorrelation is possible and it is useful as it indicates fast convergence towards true mean.

avehtari:
Excellent!

>  Other functions computed from draws (quantiles, posterior intervals, tail probabilities) may not necessarily approach the true posterior faster.

This is important point.

martinmodrak:
Ok, it's there. Turns out getting the traceplot zoom example to work on all systems is a bit of a headache :-(

martinmodrak:
OK, it turns out I was wrong - `n_eff` is explained quite well in the manual, I just never paid attention :-D. Sorry for the too quick conclusion above.

avehtari:
I had a quick look of the updated version (I guess this is the latest https://popelka.ms.mff.cuni.cz/~cerny/tmp/visual-mcmc-diagnostics-wip.html ?)

> compare_cp_ncp(mcmc_neff(neff_cp), mcmc_neff(neff_ncp), ncol = 1)

Seems to be truncating N_eff/N to 1?

> means that the mean value of the parameter computed from Stan draws approaches the true mean faster then independent samples from the true posterior 

-> means that the mean estimate of the parameter computed from Stan draws approaches the true mean faster than the mean estimate computed from independent samples from the true posterior

martinmodrak:
Reworded. The truncation at N_eff/N = 1 happens in currently released RStan and is removed in develop. I switched between develop and the CRAN version recently, so the vignette I have is with truncation. I also replaced the WIP version with the one created with the develop version of Rstan where the truncation does not happen.

avehtari:
Ok.

avehtari:
@martinmodrak great work! I approved changes, so now github let's to merge

jgabry:
These changes look good to me. Thanks! I think this is probably ready to merge. @martinmodrak anything else you wanted to add before I merge this?

martinmodrak:
Nothing. Ready to merge.

tjmahr:
I just caught up on the neff discussion. That is super interesting! I wrote the neff validation function with that assumption that neff / N could never exceed 1 because that's how it seems (intuitively).

Btw I'd like to do a run through and pull request to copyedit the vignette eventually.

jgabry:
@tjmahr sounds good. I’m sure other vignettes could use some copy editing too;)

PR #165, opened 2018-09-17
Explicitly convert `Parameter` column in melted MCMC dataframe to `factor`

### LINKED ISSUE #162: wrong parameter names bug in mcmc_areas
If users choose to pass MCMC matrix with parameter colnames as "integers string" (i.e. `colnames(mcmc_matrix) <- c("1", "2")` etc.; not a good idea to do it, but not illegal either), mcmc_areas gets very confused... or at least I think that's why I see things like this:
image

Minimal data here:
[mcmc_areas_bug.zip](https://github.com/stan-dev/bayesplot/files/2192278/mcmc_areas_bug.zip)
```r
ww <- readRDS('mcmc_areas_bug.rds')
bayesplot::mcmc_areas(ww) #polygon connecting the two
colnames(ww) <- c("a", "b")
bayesplot::mcmc_areas(ww) #all good
```

tjmahr:
closes #162

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -13,7 +13,7 @@
 * MCMC parameter names are now *always* converted to factors prior to
   plotting. We use factors so that the order of parameters in a plot matches
   the order of the parameters in the original MCMC data. This change fixes a
-  case where factor-conversion failed. (#162, @wwiecek)
+  case where factor-conversion failed. (#162, #165, @wwiecek)
 
 
 # bayesplot 1.6.0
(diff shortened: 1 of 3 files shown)


tjmahr:
The NEWS bullet describes this pull request 😄 

* MCMC parameter names are now *always* converted to factors prior to
  plotting. We use factors so that the order of parameters in a plot matches
  the order of the parameters in the original MCMC data. This change fixes a
  case where factor-conversion failed. (#162, #165, @wwiecek)

Issue #168, opened 2018-10-16
Point estimate plotted weirdly in mcmc_areas()


jtimonen:
![kernel_densities](https://user-images.githubusercontent.com/18327296/47051570-6347d700-d1ad-11e8-9aef-f77a0a152a26.png)

The point estimate (in dark blue) gets plotted like this, when I have multiple parameters to plot. Below is a code to reproduce the issue.

~~~r
require(bayesplot)
set.seed(123)

K <- 2
X <- array(0,c(N,1,K))
for(k in 1:K){
  N <- 1000
  x <- rgamma(N, 1, 5)
  X[,1,k] <- x
}

dimnames(X)[[3]] <- as.list(paste("x", 1:K, sep="_"))
mcmc_areas(X, prob_outer=0.9, point_est = "median")
~~~

 If I change median to mean, the height of the dark blue area is still off. If I change K to 1, it looks as it should. I am using version 1.6.0.

tjmahr:
I diagnosed where the flaw occurs in the bayesplot code. The problem is that each band of color is scaled to fill 95% of the vertical space, which implicitly assumed that the median and mean also corresponded to the mode.

As I document in [my blog post](https://www.tjmahr.com/ridgelines-in-bayesplot-1-5-0-release/), the mcmc areas plot is made by layering on two or three ridgeline plots. 

```r
library(bayesplot)
library(tidyverse)
theme_set(theme_grey())

set.seed(123)

K <- 2
N <- 1000
X <- array(0,c(N,1,K))
for(k in 1:K){
  x <- rgamma(N, 1, 5)
  X[,1,k] <- x
}

dimnames(X)[[3]] <- as.list(paste("x", 1:K, sep="_"))

# Separate the full range so it can be drawn on each facet.
full_range <- X %>%
  # mutate(Parameter = as.character(Parameter)) %>%
  mcmc_areas_data(prob_outer = 0.9, point_est = "median") %>%
  filter(interval == "outer") %>%
  select(-interval)

d <- mcmc_areas_data(X) %>%
  mutate(
    interval = factor(interval, c("outer", "inner", "point")))

# Separate the full range so it can be drawn on each facet.
full_range <- d %>%
  filter(interval == "outer") %>%
  select(-interval)


p_base <- ggplot(d) +
  aes(x = x, y = parameter, fill = interval, height = density) +
  # For reference include the full-range.
  # `stat = "identity"` means that ggridges should not scale the heights
[...]

Issue #163, opened 2018-08-18
support for equal-height curves mcmc_areas()


tjmahr:
A user was confused why some older `mcmc_areas()` plots changed from version 1.4.0. 

In that update, I changed `mcmc_areas()` so that all the density curves had the same area. (I *think* they have the same area.) So a wider interval would have a wide, thin density and a shorter interval is more of a spike. 

```r
library(bayesplot)
library(ggplot2)
library(dplyr, warn.conflicts = FALSE)

example_data <- bayesplot::example_mcmc_draws(params = 6)

d <- mcmc_areas_data(example_data, c("beta[1]", "beta[2]", "beta[3]")) %>%
  mutate(
    interval = factor(interval, c("outer", "inner", "point")))

ggplot(d) +
  aes(x = x, y = parameter, fill = interval, height = density) +
  ggridges::geom_density_ridges(stat = "identity", scale = 0.90) +
  scale_fill_brewer(type = "seq", palette = 2) +
  guides(fill = FALSE) +
  ggtitle("curves have same area") +
  scale_y_discrete(
    limits = unique(rev(d$parameter)),
    expand = expand_scale(mult = c(.05, .5)))
```

![image](https://user-images.githubusercontent.com/1890315/44301906-e7ccc500-a2e4-11e8-9e94-e363d053f086.png)

In the old behavior of `mcmc_areas()` the densities were scaled to have the same max height. We can easily recover this behavior.  

```r
# rescale density within each parameter to be proportion of max density
d <- d %>%
  group_by(parameter) %>%
  mutate(scaled = density / max(density)) %>%
  ungroup()

ggplot(d) +
  aes(x = x, y = parameter, fill = interval, height = scaled) +
  ggridges::geom_density_ridges(
    stat = "identity",
    scale = 0.90) +
  scale_fill_brewer(type = "seq", palette = 2) +
  guides(fill = FALSE) +
  scale_y_discrete(
    limits = unique(rev(d$parameter)),
    expand = expand_scale(mult = c(.05, .5))) +
  ggtitle("curves have same height")
```

![image](https://user-images.githubusercontent.com/1890315/44301925-4b56f280-a2e5-11e8-92e6-eeb79e0cc5c7.png)

Both styles have their benefits. The first helps compare rows. The second helps show the curvature within rows.

To fix this,

- `mcmc_areas_data()` should also return a `scaled_density` column
- `mcmc_areas()` should get an `equalize_by` argument with `"area"` `"height"` as options.
[...]

Issue #138, opened 2018-04-03
prob_outer ignored in mcmc_areas_ridges()


tjmahr:
```r
library(bayesplot)
m <- shinystan::eight_schools@posterior_sample

mcmc_areas_ridges(m, pars = "mu", regex_pars = "theta", prob_outer = .30) +
  ggplot2::ggtitle("Treatment effect on eight schools (Rubin, 1981)")
```

![image](https://user-images.githubusercontent.com/1890315/38269761-027fe6b6-3747-11e8-8817-67b83cf0ff22.png)

jgabry:
At least it's `prob_outer` and not `prob` that's broken!

PR #169, opened 2018-10-19
mcmc_areas() fixes

tjmahr:
From the NEWS.md 

* [`mcmc_areas()`](http://mc-stan.org/bayesplot/reference/MCMC-intervals.html)
  gains an argument `area_method` which controls how to draw the density 
  curves. The default `"equal area"` constrains the heights so that the curves 
  have the same area. As a result, a narrow interval will appear as a spike 
  of density, while a wide, uncertain interval is spread thin over the _x_ axis.
  Alternatively `"equal height"` will set the maximum height on each curve to 
  the same value. This works well when the intervals are about the same width. 
  Otherwise, that wide, uncertain interval will dominate the visual space 
  compared to a narrow, less uncertain interval. A compromise between the two is 
  `"scaled height"` which scales the curves from `"equal height"` using 
  `height * sqrt(height)`. (#163, #169)
  
* `mcmc_areas()` correctly plots density curves where the point estimate 
  does not include the highest point of the density curve. 
  (#168, #169, @jtimonen)

* `mcmc_areas_ridges()` draws the vertical line at *x* = 0 over the curves so 
  that it is always visible.

* `mcmc_intervals()` and `mcmc_areas()` raise a warning if `prob_outer` is ever 
  less than `prob`. It sorts these two values into the correct order. (#138)

Not in NEWS.md

* Updated all reference images in the visual unit tests.

Here are examples of the three plotting styles.

```r
library(bayesplot)
library(tidyverse)

set.seed(123)

K <- 4
N <- 1000
X <- array(0,c(N,1,K))
for(k in 1:K) {
  x <- rgamma(N, c(1, 2, 4, 5)[k], c(3, 5, 5, 6)[k])
  X[,1,k] <- x
}
X[,1,4] <- runif(1000, min = min(X), max = max(X))
dimnames(X)[[3]] <- as.list(paste("x", 1:K, sep="_"))
x <- X

p1 <- mcmc_areas(X, area_method = "equal area") + ggtitle("equal area")
p2 <- mcmc_areas(X, area_method = "equal height") + ggtitle("equal height")
p3 <- mcmc_areas(X, area_method = "scaled height") + ggtitle("scaled height")

library(patchwork)
p1 + p2 + p3 + plot_layout(ncol = 1)

```

image


tjmahr:
Closes #163. Closes #168

jgabry:
@avehtari I know you were interested in this, so just tagging you so you know it's merged now. thanks to @tjmahr the master branch of bayesplot should now have `mcmc_areas` with an argument `area_method`.

avehtari:
Great!

PR #173, opened 2019-01-08
allow negative integer input to ppc_bars, ppc_bars_grouped

### LINKED ISSUE #172: ppc_bars (and similar) throws error with negative integers
Hi, may I ask why `ppc_bars` (and other discrete outcome functions, I think) throws an error when `y` or `yrep` contain negative values? For example I would like to use this function with the [skellam distribution](https://en.wikipedia.org/wiki/Skellam_distribution) as follows:

```R
library(bayesplot)
library(skellam)

n <- 1000
m <- 500

y <- rskellam(n, 1.5, 2.5)
yrep <- matrix(rskellam(m * n, 1.4, 2.6), m, n)

ppc_bars(y, yrep)
```
which currently outputs:
```R
> ppc_bars(y, yrep)
Error in ppc_bars(y, yrep) : 
  ppc_bars expects only non-negative integers in 'y'.
```
If I remove the checks, which are:
```R
if (!all_counts(y)) 
    stop("ppc_bars expects only non-negative integers in 'y'.")
if (!all_counts(yrep)) 
    stop("ppc_bars expects only non-negative integers in 'yrep'.")
```
Then the plot works as expected:

plt

I appreciate that the majority of the time the outcomes are non-negative, but it is not always the case and I don't think that this function should error here. Thanks!

jgabry:
`ppc_bars()`, and `ppc_bars_grouped()` now check that their inputs are whole numbers rather than only positive whole numbers. 

closes #172

### DIFF
--- a/tests/testthat/test-extractors.R
+++ b/tests/testthat/test-extractors.R
@@ -4,10 +4,8 @@ context("Extractors")
 
 ITER <- 1000
 CHAINS <- 3
-capture.output(
-  fit <- stan_glm(mpg ~ wt + am, data = mtcars,
+fit <- stan_glm(mpg ~ wt + am, data = mtcars,
                   iter = ITER, chains = CHAINS, refresh = 0)
-)
 
 x <- list(cbind(a = 1:3, b = rnorm(3)), cbind(a = 1:3, b = rnorm(3)))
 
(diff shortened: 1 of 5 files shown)


tests/testthat/test-extractors.R
```diff
@@ -79,6 +79,7 @@ test_that("neff_ratio.stanreg returns correct structure", {
 test_that("rhat.stanfit returns correct structure", {
   r <- rhat(fit$stanfit)
   expect_named(r)
+  r <- r[!names(r) %in% "mean_PPD"]
```
tjmahr: This makes the tests fail now.

tests/testthat/test-extractors.R
```diff
@@ -91,6 +92,7 @@ test_that("neff_ratio.stanreg returns correct structure", {
 
   ratio <- neff_ratio(fit$stanfit)
   expect_named(ratio)
+  ratio <- ratio[!names(ratio) %in% "mean_PPD"]
```
tjmahr: This also makes the tests fail.

jgabry:
Thanks TJ. Tests are fixed so will merge this now.

PR #182, opened 2019-04-12
Remove 'arm' dependency

jgabry:
This PR removes the dependency on the 'arm' package, which was only being used internally for `ppc_error_binned()`. bayesplot now does the binning itself before plotting.


jgabry:
Great, thanks for adding that test.

PR #184, opened 2019-04-18
Call as.array on input to prepare_mcmc_array()

### LINKED ISSUE #175: Always call `as.matrix` or `as.array` on the first argument
If I have a stanreg object called `post`, then 
```
bayesplot::mcmc_acf(post)
```
should just work, but it doesn't because the first argument is neither a matrix nor an array. But when no specific method is defined for stanreg objects, the **bayesplot** functions should just call `as.matrix` or `as.array` on the first argument internally, like `apply` and things like that do. That way we can accept anything that can be coerced into a matrix or an array without having to write methods for everything.

jgabry:
Closes #175.

As suggested by @bgoodri, with this PR objects with as.array methods can now be passed to the MCMC plots (not the PPC plots) without the user having to call as.array. For example, 

```r
fit <- stan_glm(mpg ~ wt, data = mtcars)

# these give the same result now
mcmc_hist(fit)
mcmc_hist(as.array(fit))
```

If the resulting array doesn't have 2 or 3 dimensions an error is thrown.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,6 +6,9 @@
 
 <!-- Items for next release go here* -->
 
+* MCMC plots now also accept objects with an `as.array` method as 
+  input. (#175, #184)
+
 * [`mcmc_trace()`](http://mc-stan.org/bayesplot/reference/MCMC-traces.html)
   gains an argument `iter1` which can be used to label the traceplot starting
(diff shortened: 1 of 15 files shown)


jgabry:
Note: this is small PR. It says 14 files changed only because I changed the doc in one of the roxygen templates and that affects many Rd files.

PR #190, opened 2019-05-14
Add ColorBrewer palettes

### LINKED ISSUE #177: Add colorblind safe and printer friendly options to color_scheme_set
It would be nice to have colorblind safe and printer friendly options to choose with the  `color_scheme_set()` function. Options for this can be found via [here](http://colorbrewer2.org/#type=diverging&scheme=RdBu&n=6).

One option that would satisfy both critera is the color scheme `#b2182b #ef8a62 #fddbc7 #d1e5f0 #67a9cf #2166ac`.

jgabry:
Closes #177 

This PR allows ColorBrewer palettes to be used with `bayesplot::color_scheme_set()`. 

Basically,  if a user specifies `color_scheme_set("brewer-X")`, where `X` is the name of a ColorBrewer palette recognized by **RColorBrewer**, then **bayesplot** will use  `RColorBrewer::brewer.pal(n = 6, name="X")` to get the colors. 

I also cleaned up and rearranged some internals in the `bayesplot-colors.R` file and changed the roxygen to use markdown, so it may seem like more changed than actually did. The only change that changes any behavior is in these lines in the internal `scheme_from_string()` function: 

https://github.com/stan-dev/bayesplot/blob/ac6f8ae78a9018f50b3353783c91f77f659a06e6/R/bayesplot-colors.R#L251-L258

#### Example Usage

```r
color_scheme_set("brewer-Spectral")
mcmc_trace(example_mcmc_draws(), pars = "sigma")
```
brewer-example

### DIFF
--- a/R/bayesplot-colors.R
+++ b/R/bayesplot-colors.R
@@ -227,61 +227,6 @@ plot_scheme <- function(scheme) {
     )
 }
 
-
-# @param scheme A string (length 1) naming a scheme
-scheme_from_string <- function(scheme) {
-  stopifnot(length(scheme) == 1)
-  if (identical(substr(scheme, 1, 4), "mix-")) {
-    to_mix <- unlist(strsplit(scheme, split = "-"))[2:3]
-    x <- setNames(mixed_scheme(to_mix[1], to_mix[2]), scheme_level_names())
-    return(structure(x, mixed = TRUE, scheme_name = scheme))
-  } else if (identical(substr(scheme, 1, 7), "brewer-")) {
-    if (!requireNamespace("RColorBrewer", quietly = TRUE)) {
-      stop("Please install the 'RColorBrewer' package to use a ColorBrewer scheme.",
-           call.=FALSE)
-    }
-    clrs <- RColorBrewer::brewer.pal(n = 6, name = gsub("brewer-", "", scheme))
-    x <- setNames(as.list(clrs), scheme_level_names())
-    return(structure(x, mixed = FALSE, scheme_name = scheme))
-  } else {
-    scheme <- match.arg(scheme, choices = names(master_color_list))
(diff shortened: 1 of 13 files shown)


R/bayesplot-colors.R
```diff
@@ -310,34 +357,8 @@ is_hex_color <- function(x) {
   isTRUE(nchar(x) == 7)
```
tjmahr, line 366: ggplot2 will take nine character ones where the last two set the alpha level. Some palettes emit colors with these extra characters.

```
scales::viridis_pal()(4)
#> [1] "#440154FF" "#31688EFF" "#35B779FF" "#FDE725FF"
```

Not sure we need to accommodate this though.

R/bayesplot-colors.R
```diff
@@ -259,6 +314,7 @@ get_color <- function(levels) {
   color_vals <- color_scheme_get()[levels]
   unlist(color_vals, use.names = FALSE)
 }
+
 full_level_name <- function(x) {
```
tjmahr, line 315: I'm adding a commit where I refactor this to not use a `switch()` but use the named-vector look-up table trick to get vectorization for free. http://adv-r.had.co.nz/Subsetting.html#lookup-tables

tjmahr:
Take a look at my changes and feel free to merge.

R/bayesplot-colors.R
jgabry: Oh, I didn't know that. I'm not sure we need to accommodate it either, although I'm not opposed. I suppose adding support for that can be a separate issue if we every want to do it or get a request.

jgabry:
Thanks @tjmahr. Merging this now. (Travis checks are good except for a note about the hexbin package but that’s taken care of by my other PR https://github.com/stan-dev/bayesplot/pull/187)

PR #192, opened 2019-05-16
Use markdown in roxygen

### LINKED ISSUE #191: Use markdown in roxygen doc
I converted all of the doc for the **loo** package recently to use markdown syntax in the roxygen doc (https://github.com/stan-dev/loo/pull/108) and I think it's worth doing for **bayesplot**. The [roxygen2md](https://github.com/r-lib/roxygen2md) helps but then there's a bunch of manual editing required afterwards. I'll do it when I have time, but in the meantime we can also start the process by changing individual files that we're editing for other reasons (e.g., #190 does this for the `bayesplot-colors.R` file).

jgabry:
closes #191 

So much for doing this gradually... This PR converts _all_ roxygen doc to use markdown syntax where possible. There are a few cases where I've left the traditional syntax in place because there are no good alternatives to lists made with `\describe{...}` that I'm aware of (it does more than just what you get from markdown lists)  and I don't think latex for math is supported without still using `\eqn{}`. 

@tjmahr Do you want to review this or should I just merge it? From what I can tell all the man pages still look good and it could be a pain to actually review it. Up to you.


man/theme_default.Rd
```diff
@@ -2,26 +2,24 @@
...
   base_family = getOption("bayesplot.base_family", "serif"))
 }
 \arguments{
 \item{base_size, base_family}{Base font size and family (passed to
-\code{\link[ggplot2]{theme_bw}}). It is possible to set
-\code{"bayesplot.base_size"} and \code{"bayesplot.base_family"} via
-\code{\link{options}} to change the defaults, which are \code{12} and
-\code{"serif"}, respectively.}
+\code{\link[ggplot2:theme_bw]{ggplot2::theme_bw()}}). It is possible to set \code{"bayesplot.base_size"} and
+\code{"bayesplot.base_family"} via \code{\link[=options]{options()}} to change the defaults, w
```
tjmahr: which split across lines

man/bayesplot-package.Rd
```diff
@@ -4,60 +4,57 @@
...
 packages for Bayesian modeling, particularly (but not necessarily) packages
-powered by \pkg{\link[rstan]{rstan}}. Examples of packages that will soon (or
-already are) using \pkg{bayesplot} are \pkg{rstan} itself, as well as the
-\pkg{rstan}-dependent \pkg{rstanarm} and \pkg{brms} packages for applied
-regression modeling.
+powered by \link[rstan:rstan-package]{rstan} (the \R interface to Stan).
+Examples of packages that will soon (or already are) using \strong{bayesplot} are
+\strong{rstan} itself, as well as the \strong{rstan}-dependent \strong{rstanarm} and
+\strong{brms} packages for applied regression modeling.
 }
```
tjmahr, line 24: This is just an observation... I'm noticing that all the \pkg tags turn into \strong codes. I guess that will cause them to turn bold on the pkgdown site too, which is fine.

man/theme_default.Rd
tjmahr: i fixed it

man/bayesplot-package.Rd
jgabry: That's actually something that I did manually. roxygen2md leaves them as `\pkg{pkgname}` in the roxygen doc, but having `\pkg{pkgname}` in an Rd file has the effect of making the package name bolded when viewing the doc anyway so I just switched it to `**pkgname**`.

jgabry:
@tjmahr Thanks! Merging now.

PR #219, opened 2020-04-28
Fixes for dplyr 1.0.0

hadley:
The main change is that `summarise()` no longer strips the names off of columns, so you have to do that yourself.

I also fixed a few warnings to make sure I wasn't missing anything in the tests. I hope that's ok!


jgabry:
Thanks @hadley!

jgabry:
Everything passes locally for me so will merge. This is failing on travis but only because of issues installing rstan apparently.

Issue #218, opened 2020-04-06
remove extra space from mcmc_areas


avehtari:
I needed a mcmc_areas plot for a book, and noticed that all mcmc_areas plots have unnecessary extra space on top and bottom. This is not clear with just a few variables, but with 4 than more it's visible e.g. in bodyfat, candy, diabetes, mesquite, roaches, and winequality case studies at https://avehtari.github.io/modelselection/
It gets worse and worse with more variables.
For example this kind of call has empty space
```
mcmc_areas(as.matrix(fit), pars=vars(-'(Intercept)',-sigma))
```
and it can be fixed with
```
p0 <- mcmc_areas(as.matrix(fit0), pars=vars(-'(Intercept)',-sigma)
p0 <- p0 + scale_y_discrete(limits = rev(levels(p0$data$parameter)))
p0
```
This approach is a bit annoying, so I propose to make the plots tight by default.

This had slipped past previously as in notebooks there is plenty of space, but for a book I need tight layout and then I noticed that the extra space is actually appearing in many plots.

jgabry:
@avehtari Thanks for bringing this up and sorry for the slow response. I finally made some time recently to dig back into bayesplot issues. 

This was a bit tricky because if we don't add the `expand` argument to `scale_y_discrete()` then sometimes the tops of the areas get slightly cut off, but the settings we were using definitely created way too much space on top. 

@tjmahr might have a better idea, but after playing around with a bunch of options and trying with different numbers of parameters, what ended up working best (I think) was to use an additive expansion that decreases with the number of parameters. I put this on a branch `remove-extra-space-mcmc_areas`. Here's a comparison with the current master branch: 

For 2, 4, 10, and 20 parameters (simulated from standard normals) here is what the current master branch produces: 
![areas_old](https://user-images.githubusercontent.com/7796803/85243932-2bc50f80-b411-11ea-8312-fe4a07247a70.png)

and here is what the version on the new branch `remove-extra-space-mcmc_areas` looks like for the same parameters:

![areas_new](https://user-images.githubusercontent.com/7796803/85243975-51521900-b411-11ea-8b24-e41e5c90d470.png)

What do you think?

avehtari:
New version looks much better. It would be useful to add in the documentation a comment how to adjust the spacing manually, if even more tight spacing is desired.

jgabry:
Good idea, I'll add an example in the Examples section of the doc

PR #224, opened 2020-06-18
allow changing size of ridgelines

jgabry:
Based on this issue reported on the forums 

https://discourse.mc-stan.org/t/bayesplot-mcmc-areas-line-and-grid-width-manipulation/16001

I added a `size` argument to `mcmc_areas()` and `mcmc_areas_ridges()` to control the size (thickness) of the ridgelines.

### DIFF
--- a/R/mcmc-intervals.R
+++ b/R/mcmc-intervals.R
@@ -245,6 +245,8 @@ mcmc_intervals <- function(x,
 
 #' @rdname MCMC-intervals
 #' @export
+#' @param size For `mcmc_areas()` and `mcmc_areas_ridges()`, the size of the
+#'   ridgelines.
 mcmc_areas <- function(x,
                        pars = character(),
                        regex_pars = character(),
@@ -255,6 +257,7 @@ mcmc_areas <- function(x,
                        prob_outer = 1,
                        point_est = c("median", "mean", "none"),
                        rhat = numeric(),
+                       size = NULL,
                        bw = NULL,
                        adjust = NULL,
                        kernel = NULL,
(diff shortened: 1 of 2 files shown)


jgabry:
Travis still isn't working because it times out, but tests pass locally for me.

jgabry:
All checks now passing via GitHub actions

jgabry:
Added visual tests and made sure axis expansion is sufficient to avoid cutting off top of the highest ridgeline (I checked plots with few parameters, medium number of parameters, and many parameters).

R/mcmc-intervals.R
```diff
@@ -291,6 +293,7 @@ mcmc_areas <- function(x,
                        prob_outer = 1,
                        point_est = c("median", "mean", "none"),
                        rhat = numeric(),
+                       size = NULL,
```
tjmahr: should we have a more descriptive name than size? "border_size"?

R/mcmc-intervals.R
```diff
@@ -522,14 +535,21 @@ mcmc_areas_ridges <- function(x,
...
+      args_inner$size <- size
+    }
+
     layer_list_inner[[par_num]] <- do.call(ggridges::geom_ridgeline, args_inner)
   }
 
   ggplot(datas$outer) +
     aes_(x = ~ x, y = ~ parameter) +
     layer_outer +
     scale_y_discrete(limits = unique(rev(data$parameter)),
```
tjmahr, line 549: is this compatible with "mcmc_areas() tries to use less vertical blank space. (#218, #230)?" 

I'm talking about the `scale_y_discrete()` but it's cut off here.

tjmahr (commented):
Just a couple of questions. Otherwise looks good to me.

R/mcmc-intervals.R
jgabry: Good call. Changing to `border_size`.

R/mcmc-intervals.R
jgabry: Will double check before merging this

PR #226, opened 2020-06-18
Use GitHub Actions for CI

rok-cesnovar:
Replaces Travis with Github Actions. This is the default script made with 

`usethis::use_github_action_check_full()`

and the `codecov` example from r-lib https://github.com/r-lib/actions/tree/master/examples#test-coverage-workflow

Windows tests with 3.6 are not included as I think there is an issue with backsports package or something. Not sure its relevant for bayesplot.


jgabry:
Thanks Rok! I think this should work. I pushed a commit to fix the test coverage failure and I think one tweak I made to the vignette should fix ubuntu failures. I'll let the checks run again before merging.

rok-cesnovar:
That did it yeah. Great that it worked. The macOS devel one takes a while the first time (~0.5hour) then it gets cached. I know you also like the [skip ci] in commit messages thing, so might also want to add that.

jgabry:
> I know you also like the [skip ci] in commit messages thing, so might also want to add that.

Yeah I like that so we can skip running all the tests if all we're doing is something like editing the readme file. Will add it.

PR #227, opened 2020-06-20
add extractor methods for CmdStanMCMC objects (from CmdStanR)

jgabry:
Adds extractor functions (`nuts_params`, `neff_ratio`, etc.) for models fit using CmdStanR. Also added tests, but they are skipped on CRAN and if CmdStanR is not installed.


jgabry:
Thanks, yeah I forgot to update NEWS. Just added a news entry.

PR #230, opened 2020-06-22
Remove extra space in mcmc_areas

### LINKED ISSUE #218: remove extra space from mcmc_areas

jgabry:
Fixes #218

Copying from my comment in #218: 

after playing around with a bunch of options and trying with different numbers of parameters, what ended up working best (I think) was to use an additive expansion that decreases with the number of parameters. I put this on a branch `remove-extra-space-mcmc_areas`. Here's a comparison with the current master branch: 

For 2, 4, 10, and 20 parameters (simulated from standard normals) here is what the current master branch produces: 
areas_old

and here is what the version in this PR looks like for the same parameters:

areas_new


tjmahr:
While we are under the hood for this one, I'd like to fix the clipping where the line in the top panel gets cut off.  

image

The demos you supply here appear to fix this problem, but I can still get it if I use `area_method = "equal areas"` in this version. 

I'm trying the `geom_blank()` strategies of adjusting plotting area by plotting invisible data.

jgabry:
> While we are under the hood for this one, I'd like to fix the clipping where the line in the top panel gets cut off.

Good call, thanks. 

> Can you try out your diagnostic plots using the updated version?

Yup, will check now.

jgabry:
> Can you try out your diagnostic plots using the updated version?

Do you mean the same set of plots I made above? I made them again (although I forgot to set a seed last time so they'll look a bit different). They still look good after your change. Here's using default arguments: 

areas_new-default

And here's using `prob = 2/3, prob_outer = 0.9, point_est = "mean"`: 

areas_new-default-2

Issue #233, opened 2020-07-28
PPC for (right-)censored data


fweber144:
@paul-buerkner asked for an issue here on bayesplot's issue tracker concerning [brms's issue #967](https://github.com/paul-buerkner/brms/issues/967). So here it is. There's also a corresponding thread in [The Stan Forums](https://discourse.mc-stan.org/t/ppc-for-right-censored-data/17032).

That brms issue is a feature request for a PPC for right-censored data in parametric time-to-event models with a response distribution from which sampling is easy (e.g. log-normal or Weibull models). The suggestion is to compare the Kaplan-Meier estimate of the CCDF for the observed data to the posterior CCDFs. Section 4.4 ("Standardised survival probabilities") of the [preprint by Brilleman et al. (2020)](https://arxiv.org/abs/2002.09633) describes a similar PPC, however averaging over individual posterior CCDFs. This averaging over individual posterior CCDFs should not be necessary when sampling from the response distribution, which makes the approach suitable for brms's generic PPC framework. @jgabry suggested to integrate such a PPC into the bayesplot package where it may be accessed by other packages as well.

You may find some example code in [brms's issue #967](https://github.com/paul-buerkner/brms/issues/967).

jgabry:
Thanks for opening the issue. I've been a bit swamped recently so I haven't had much time to think about this yet, but I definitely think it would be good to have PPCs for this in bayesplot. I'll try to respond with more soon!

fweber144:
No problem, I don't have that much time either. But if it helps, I could try a PR. For this, it would be necessary to decide if you want to include this feature in the existing `bayesplot::ppc_ecdf_overlay()` function or if you want to create a new function for that. Including it in `bayesplot::ppc_ecdf_overlay()` will probably require a new argument accepting the event/censoring indicator and (if you want to use the code provided [here](https://github.com/paul-buerkner/brms/issues/967#issuecomment-663032952)) a back-transformation of the CCDFs to CDFs via "CDF = 1 - CCDF". I guess this back-transformation shouldn't be too difficult with ggplot2; I will check it out if I get the time.

fweber144:
[...]

PR #234, opened 2020-08-12
ppc_km_overlay(): PPC for (right-)censored data

fweber144:
This PR adds the feature requested in #233 by adding the function `ppc_km_overlay()`. Notes:
* I have only considered right censoring.
* For argument `status_y` of `ppc_km_overlay()`, I used survival's/rstanarm's convention of an *event* indicator (see `?survival::Surv` and the [preprint by Brilleman et al. (2020)](https://arxiv.org/abs/2002.09633), section 2.1). An alternative would be brms's convention of a *censoring* indicator (see `?brms::brmsformula`), but I think survival's/rstanarm's convention of an *event* indicator is more common.
* Due to the code from `validate_group()`, argument `status_y` must have more than one unique value. However, this shouldn't raise any objections: If there are no censored observations, the Kaplan-Meier estimator reduces to "1 - ECDF", so the user could use `ppc_ecdf_overlay()`. If there are no events (i.e. only right-censored observations), then the survival curve is just a horizontal line at height 1 which is probably not of any value in practice.
* I thought about allowing censored *replicated* observations, but I think this would lead to more confusion than it would help. And besides, I don't see why it would ever make sense to censor the replicated observations.
* I tried to stick as close as possible to `ppc_ecdf_overlay()`, with one exception: In `ppc_km_overlay()`, I hard-coded the use of `geom_step()`. In contrast, `ppc_ecdf_overlay()` has an argument `discrete` which defaults to `FALSE` so that `geom = "line"` is used. I am using `geom_step()` in `ppc_km_overlay()` because in my opinion, the ECDF (and therefore also the ECCDF and the Kaplan-Meier estimate) is *defined* to be a step function. If you want, I can open a new issue for this as far as `ppc_ecdf_overlay()` is concerned.
* With `devtools::check()`, I get the NOTE `release_questions: missing arguments not allowed in calls to 'c'`. This seems to be related to https://github.com/stan-dev/bayesplot/blob/52f1a9f48a2672c512cfcc02721a6e2b3197649a/R/bayesplot-package.R#L99-L103


jgabry:
Awesome, thanks for the PR! 

Just a heads up that I may be a bit delayed in fully reviewing this because StanCon is tomorrow and then I'm giving a talk this weekend at the (virtual) NY R conference. Until then a few comments on your bullet points: 


> * I have only considered right censoring.

I think that's fine since that's the most common form of censoring in survival data (right?). Do you think for left censoring and interval censoring a different type of plot would make more sense or could support for those be added to this function in the future? 

> * For argument `status_y` of `ppc_km_overlay()`, I used survival's/rstanarm's convention of an _event_ indicator (see `?survival::Surv` and the [preprint by Brilleman et al. (2020)](https://arxiv.org/abs/2002.09633), section 2.1). An alternative would be brms's convention of a _censoring_ indicator (see `?brms::brmsformula`), but I think survival's/rstanarm's convention of an _event_ indicator is more common.

Sounds good. 

> * Due to the code from `validate_group()`, argument `status_y` must have more than one unique value. However, this shouldn't raise any objections: If there are no censored observations, the Kaplan-Meier estimator reduces to "1 - ECDF", so the user could use `ppc_ecdf_overlay()`. If there are no events (i.e. only right-censored observations), then the survival curve is just a horizontal line at height 1 which is probably not of any value in practice.

Ok that's fine. If you didn't already can you mention this in the doc? Alternatively, I think it would be fine to add an argument to `validate_group()` telling it whether or not to error if there's not more than 1 unique value. 

> * I thought about allowing censored _replicated_ observations, but I think this would lead to more confusion than it would help. And besides, I don't see why it would ever make sense to censor the replicated observations.

Yeah I agree. If someone comes up with a good use case for this we can revisit it in the future, but I don't see why it would make sense either. 

> * I tried to stick as close as possible to `ppc_ecdf_overlay()`, with one exception: In `ppc_km_overlay()`, I hard-coded the use of `geom_step()`. In contrast, `ppc_ecdf_overlay()` has an argument `discrete` which defaults to `FALSE` so that `geom = "line"` is used. I am using `geom_step()` in `ppc_km_overlay()` because in my opinion, the ECDF (and therefore also the ECCDF and the Kaplan-Meier estimate) is _defined_ to be a step function. If you want, I can open a new issue for this as far as `ppc_ecdf_overlay()` is concerned.

To be honest, right now I forget why we did that for `ppc_ecdf_overlay()` (I'll take a look through the issues and PRs at some point and see if I can track it down). I think it's fine to just use `geom_step()` for the new function.

> * With `devtools::check()`, I get the NOTE `release_questions: missing arguments not allowed in calls to 'c'`. This seems to be related to https://github.com/stan-dev/bayesplot/blob/52f1a9f48a2672c512cfcc02721a6e2b3197649a/R/bayesplot-package.R#L99-L103

Yeah you're right, looks like a comma that shouldn't be there. If you want you can remove the comma but otherwise don't worry about that (we can deal with it separately). 


Anyway, thanks again for the PR! This will be a nice addition. I'll try to review the code as soon as possible.

fweber144:
Sorry - I'm quite busy at the moment, but I'll respond as soon as I can.

jgabry:
No problem, I've been busy too! This reminds me that I still need to give this a thorough review.

fweber144:
I got the opportunity to work on this sooner as I thought, so here are my replies:

> 
> > * I have only considered right censoring.
> 
> I think that's fine since that's the most common form of censoring in survival data (right?). Do you think for left censoring and interval censoring a different type of plot would make more sense or could support for those be added to this function in the future?
> 

Yes, right censoring is the most common form of censoring in survival analyses.

In principle, I think this type of plot (estimated CCDF) should also be appropriate for left and interval censoring, but the estimated CDF should be appropriate as well. This should mainly be a question of convention. I haven't dealt with left or interval censoring in an own analysis yet, so I have no experience concerning this convention. Personally, I would choose the CCDF as it is more consistent with the convention for right-censored data. The [NADA](https://cran.r-project.org/web/packages/NADA/index.html) package provides the function `cenfit()` which gives a KM-type estimate for the *CDF* in case of left-censored data. So perhaps the CDF is also common for left-censored data. But regardless of the type of plot, argument `status_y` of `ppc_km_overlay()` may be adopted to support non-right censoring in the future. However, if the type of plot then differs from the CCDF (e.g. because the CDF is plotted), then the name `ppc_km_overlay()` might be misleading. But I can't think of a much better name. Something like `ppc_cens_overlay()` is not specific for the resulting plot (for example, one might also think that this would produce a density estimate for censored data, even though I don't know if that exists). So I would stick with `ppc_km_overlay()` and instead adopt the resulting plot (if necessary). For example, if you decide to use NADA for left-censored data, then I think it would be best to transform the resulting CDF to a CCDF (and keep the name `ppc_km_overlay()`). But if you have a better name for `ppc_km_overlay()`, I have no problem with renaming it.
> 
> > * Due to the code from `validate_group()`, argument `status_y` must have more than one unique value. However, this shouldn't raise any objections: If there are no censored observations, the Kaplan-Meier estimator reduces to "1 - ECDF", so the user could use `ppc_ecdf_overlay()`. If there are no events (i.e. only right-censored observations), then the survival curve is just a horizontal line at height 1 which is probably not of any value in practice.
> 
> Ok that's fine. If you didn't already can you mention this in the doc? Alternatively, I think it would be fine to add an argument to `validate_group()` telling it whether or not to error if there's not more than 1 unique value.
>
I followed your suggestion to add an argument to `validate_group()` telling it whether or not to error if there is only one unique value. The PR should have been updated accordingly.
> 
> > * With `devtools::check()`, I get the NOTE `release_questions: missing arguments not allowed in calls to 'c'`. This seems to be related to https://github.com/stan-dev/bayesplot/blob/52f1a9f48a2672c512cfcc02721a6e2b3197649a/R/bayesplot-package.R#L99-L103
> 
> Yeah you're right, looks like a comma that shouldn't be there. If you want you can remove the comma but otherwise don't worry about that (we can deal with it separately).
> 
I removed that comma now. The PR should have been updated accordingly.

jgabry:
Sorry for the delay on this! Thanks again for working on this, it looks great. I just have a few comments/questions: 

* After thinking about it some more, I realized that we can just get rid of the check for unique `group` values entirely instead of turning it off in particular cases. The plotting functions that use the group variable will still work if there's only 1 unique group value, they'll just create 1 facet. So we might as well not error there. That avoids requiring any special behavior for this new function. Sorry for asking you to add that argument to check unique values! I just pushed a commit to fix this so no action necessary on your part.

* I didn't realize this before but the documentation at `?ggplot2::fortify` says that it might be deprecated in the future and that the broom package is recommended instead. Right now there's no deprecation warning but I wonder if we should avoid using it so that we don't need to fix this in the future when they deprecate it. What do you think?

* Currently the function is documented on the `PPC-distributions` page, which makes sense since this is similar to `ppc_ecdf_overlay`. However, if we add more functions for plotting survival models at some point then it probably makes sense to have a separate documentation page for survival model plots (e.g. a `PPC-survival` page). So I'm tempted to start that separate documentation page now. What do you think? 

* The way the code is right now `y` is actually plotted underneath `yrep` but I think it should be reversed. Here's an example: 

```r
data(kidney, package = "brms")
kidney_fit <- brm(
  formula = time | cens(censored) ~ age * sex + disease + (1|patient),
  data = kidney,
  family = weibull()
)
ppc_km_overlay(
  y = kidney$time, 
  yrep = posterior_predict(kidney_fit, nsamples = 50), 
  status_y = ifelse(kidney$censored == 1, 0, 1) # need to convert censored to status
)
```  


So the lines for `yrep` are covering up the line for `y`. 


Ok, that's all the comments/questions that I have. Aside from these minor issues I think this is ready to go. Thanks again for working on it and sorry again for taking so long to get back to you!

fweber144:
> 
>   * I didn't realize this before but the documentation at `?ggplot2::fortify` says that it might be deprecated in the future and that the broom package is recommended instead. Right now there's no deprecation warning but I wonder if we should avoid using it so that we don't need to fix this in the future when they deprecate it. What do you think?
> 

I didn't know that `ggplot2::fortify()` may get deprecated. *Ad hoc*, I can think of the following solutions:

* Use the `ggfortify:::fortify.survfit()` method directly instead of the `ggplot2::fortify()` generic.
* Use `autoplot()` (i.e. the `ggplot2::autoplot()` generic which redirects to the `ggfortify::autoplot.survfit()` method) and extract the data from the returned `ggplot` object. Then however, my note from [this comment](https://github.com/stan-dev/bayesplot/issues/233#issuecomment-668465600) has to be kept in mind:
  
  > Note that in contrast to `ggfortify::autoplot.survfit()`, `ggfortify::fortify.survfit()` does not add the survival proportion 1 at timepoint 0 by default. Sooner or later, adding the survival proportion 1 at timepoint 0 might cause problems in the general context of bayesplot. Taking the more conservative default of `ggfortify::fortify.survfit()` (not adding the survival proportion 1 at timepoint 0) would be preferable for bayesplot, I think.
  
  But that issue could most probably be circumvented by manually deleting the added survival proportion 1 at timepoint 0 (but only if it was added automatically).

Nevertheless, I would wait with these alternative solutions since according to its documentation, `ggplot2::fortify()` *may* be deprecated (so this does not seem to be a final decision).

>   * Currently the function is documented on the `PPC-distributions` page, which makes sense since this is similar to `ppc_ecdf_overlay`. However, if we add more functions for plotting survival models at some point then it probably makes sense to have a separate documentation page for survival model plots (e.g. a `PPC-survival` page). So I'm tempted to start that separate documentation page now. What do you think?
> 

Yes, creating a separate documentation page makes sense if you plan to create other PPC plots for censored data. But perhaps a more general name for this page would be better? Something like `PPC-censoring` instead of  `PPC-survival`? Because censoring not only occurs in survival analysis.

> 
>   * The way the code is right now `y` is actually plotted underneath `yrep` but I think it should be reversed. Here's an example:
> 
> 
> ```r
> data(kidney, package = "brms")
> kidney_fit <- brm(
>   formula = time | cens(censored) ~ age * sex + disease + (1|patient),
>   data = kidney,
>   family = weibull()
> )
> ppc_km_overlay(
>   y = kidney$time, 
>   yrep = posterior_predict(kidney_fit, nsamples = 50), 
>   status_y = ifelse(kidney$censored == 1, 0, 1) # need to convert censored to status
> )
> ```
> 
> 
> 
> So the lines for `yrep` are covering up the line for `y`.
> 

I solved that issue and pushed the changes. Note that I also updated the Rd documentation and the tests. For the tests, I simply removed those which were not up-to-date anymore (since argument `group` of `validate_group()` is now allowed to have a single unique value). So I didn't write any new tests to replace those which I removed.

jgabry:
Sorry for the slow reply, been super busy! 

> * Use the `ggfortify:::fortify.survfit()` method directly instead of the `ggplot2::fortify()` generic.

I wish they exported `fortify.survfit()` because calling anything with three colons `:::` will get flagged by CRAN. So we can't actually call that function directly. 

> Nevertheless, I would wait with these alternative solutions since according to its documentation, `ggplot2::fortify()` _may_ be deprecated (so this does not seem to be a final decision).

Yeah since there are workarounds we know we can use I think it's fine to wait to address this until it's necessary. 

> Something like `PPC-censoring` instead of `PPC-survival`? Because censoring not only occurs in survival analysis.

Good point. Do you want to move the doc to a new `PPC-censoring` page as part of this PR? But no worries if you don't have time to do that, I can always do it later if we really want it.

jgabry:
Also I just pushed a commit adding an item to the NEWS file and I added you as a contributor in the DESCRIPTION file. I assume that's ok, but if you'd rather not be listed let me know. 

I'm going to mark this PR as approved because I think everything has been addressed. Will wait to merge in case you want to add the separate `PPC-censoring` documentation page, which would be great but it's not strictly necessary so I'm happy to merge this without that if you don't have time to do it.

jgabry (approved):

fweber144:
> Good point. Do you want to move the doc to a new PPC-censoring page as part of this PR? But no worries if you don't have time to do that, I can always do it later if we really want it.

Yep, done. I hope the example is ok this way. Note that for the references, I removed "reference-vis-paper", kept "reference-bda", and added "reference-km" (a reference for Kaplan and Meier, 1958). Change that if you want. For Kaplan and Meier (1958), I first followed the citation style from "reference-vis-paper", but then I followed the citation style from "reference-loo" (commit 3836da1). You can decide which citation style to use.

> Also I just pushed a commit adding an item to the NEWS file and I added you as a contributor in the DESCRIPTION file. I assume that's ok, but if you'd rather not be listed let me know.

Thank you, that's ok.

fweber144:
Sorry, I forgot to add a link on the `PPC-overview` page. I will fix that asap.

fweber144:
> Sorry, I forgot to add a link on the `PPC-overview` page. I will fix that asap.

Now fixed.

jgabry:
This looks great, thanks a lot! Merging now.

fweber144:
Great, thanks! I closed #233.

Issue #171, opened 2018-11-02
Fixing the density estimation for ppc_loo_pit_overlay


GustafRydevik:
The tendency of the density estimation to go down at the edges can cause confusing graphs, especially for situations where the data set is small. Would it be possible to change this to use a boundary-corrected kernel density estimator, such as kde.boundary in  https://cran.r-project.org/web/packages/ks/ or bckden in https://rdrr.io/cran/evmix/man/bckden.html ?

avehtari:
This has been on our todo list, but not with high priority. If you have tried either of these function you propose, it would be great if you can include an example of how to use them and how the plot would look like for an uniform data.

PR #235, opened 2020-08-13
Boundary corrected KDE values and flag: ppc_loo_pit_overlay()

ecoronado92:
@jgabry @avehtari Included a `bc_kde` optional flag to `ppc_loo_pit_overlay` for users to decide whether to compute the boundary corrected densities or the current `stat_density` ones based on #171  

Flag is set to `FALSE` by default, and appropriate documentation has been added to function params and man stating the following

```rmd
For `ppc_loo_pit_overlay()`, when set to TRUE function will
compute `"boundary corrected"` density values using beta kernel estimators.
Although preferable, the current implementation is slow (default = FALSE).
```


jgabry:
Awesome, thanks for this!! Will review as soon as possible.

R/ppc-loo.R
```diff
@@ -117,6 +119,9 @@ NULL
 #'   standard normal distribution.
 #' @param trim Passed to [ggplot2::stat_density()].
 #' @template args-density-controls
+#' @param bc_kde For `ppc_loo_pit_overlay()`, when set to TRUE function will
+#'    compute `"boundary corrected"` density values using beta kernel estimators.
+#'    Although preferable, the current implementation is slow (default = FALSE).
```
jgabry:
```suggestion
#' @param bc_kde For `ppc_loo_pit_overlay()`, when set to `TRUE` the function will
#'    compute boundary corrected density values using beta kernel estimators.
#'    Although we recommend using this option, the current implementation is slow 
#'    so it is set to `FALSE` by default.
```

Also should mention that some of the arguments for controlling the kernel density estimation are ignored if this is TRUE.

R/ppc-loo.R
```diff
@@ -144,50 +150,78 @@ ppc_loo_pit_overlay <- function(y,
...
-    yaxis_text(FALSE) +
-    yaxis_ticks(FALSE)
+  
+ 
+  if (bc_kde){
+    pit <- .bc_pvals(x = pit, bw = bw)
+    unifs <- t(apply(unifs, 1, function(x) .bc_pvals(x, bw = bw)))
+    
+    data <- ppc_data(pit, unifs) %>% 
+      dplyr::arrange(rep_id) %>%
```
jgabry:
```suggestion
      dplyr::arrange(.data$rep_id) %>% 
```

This would get flagged when CRAN runs its checks because it won't see that `rep_id` exists anywhere. In package source code we can't use the non-standard evaluation that is typically used with dplyr in interactive use so we have to use the slightly uglier version.

R/ppc-loo.R
```diff
@@ -456,3 +490,81 @@ ppc_loo_ribbon <-
...
+    warning(paste("Ignored", sum(valid_pvals == 0),
+                  "PIT values == 0, they are invalid for beta KDE method"))
+    valid_pvals = valid_pvals[valid_pvals != 0]
+  }
+  
+  # Boundary corrected KDE values
+  bc_vals <- .bc_dunif(xs = xs, pvals=valid_pvals, b =bw)
+  
+  # Set any negative values to zero and output bc density values
+  bc_vals[which(bc_vals < 0)] = 0
```
jgabry:
```suggestion
  bc_vals[which(bc_vals < 0)] <- 0
```

R/ppc-loo.R
```diff
@@ -456,3 +490,81 @@ ppc_loo_ribbon <-
...
+  d[ifelse(!is.na(xs), (xs >= 0) & (xs <= 1), FALSE)] = bc_vals
```
jgabry:
```suggestion
  d[ifelse(!is.na(xs), (xs >= 0) & (xs <= 1), FALSE)] <- bc_vals
```

jgabry: Also, are we always guaranteed that `bc_vals` will be exactly the same length as the number of `TRUE`s resulting from the ifelse? If yes then this is fine. If not then we could conceivably get an error here in some cases.

R/ppc-loo.R
```diff
@@ -456,3 +490,81 @@ ppc_loo_ribbon <-
...
+  d <- xs
+  xmax <- 1 + 1e-9
+  bw <- density(x, bw = bw)$bw # extract bw
+  
+  # get only valid pvals
+  valid_pvals <- x[is.finite(x)]
+  
+  # Some sanity checks
+  if (abs(xmax - max(valid_pvals)) >= 1e-1) {
+    stop("largest PIT value must be below 1")
```
jgabry: I like the sanity check, but this seems really strict. The first example I tried (the rstanarm model from https://github.com/stan-dev/bayesplot/blob/master/tests/testthat/test-ppc-loo.R) caused this to error. But if I disable this sanity check the plot seems to work fine for that example.

R/ppc-loo.R
ecoronado92: Thanks for all the comments! Agreed, I'll make this change.

R/ppc-loo.R
ecoronado92: Same as above, I'll change this accordingly

R/ppc-loo.R
ecoronado92: My bad haha, I'll make sure style is consistent. I've been moving between R and Python recently, so some pythonic habits turn over.

R/ppc-loo.R
ecoronado92: Actually, you bring up a good point. I'll make some change that should take that into account. This is guaranteed for the reference uniforms (given these are generated via `runif` but could be a problem for the PIT values.

I'll make some changes accordingly

R/ppc-loo.R
ecoronado92: This was my attempt at floating point numbers, for some when I was comparing 1.000000000 to xmax = 1 I'd get this to stop which was an issue

ecoronado92:
@jgabry Hey! Thanks for the comments. I've made some changes and added some comments to your points above.

I also tried using the https://github.com/stan-dev/bayesplot/blob/master/tests/testthat/test-ppc-loo.R and didn't get an error. Would you mind checking if the error persists with the new changes?

jgabry:
Thanks for the changes! They look good. Just one thing: 

> I also tried using the https://github.com/stan-dev/bayesplot/blob/master/tests/testthat/test-ppc-loo.R and didn't get an error. Would you mind checking if the error persists with the new changes?

Just to clarify, I wasn't getting an error running those tests, I was getting an error using the `stan_glm` model from those tests but setting `bc_kde=TRUE`. Now if I do that with `boundary_correction=TRUE` I don't get an error but I get a bunch of warnings of the form: 

```
50: In if (valid_pvals > xmax) { ... :
  the condition has length > 1 and only the first element will be used
```

because `if` doesn't like that `valid_pvals` is a vector. I just pushed a test in https://github.com/stan-dev/bayesplot/pull/235/commits/ed455bc5d17d40e45c2da728de5eae5f13df8271 to catch this, which will currently fail because of those warnings.

ecoronado92:
```
50: In if (valid_pvals > xmax) { ... :
  the condition has length > 1 and only the first element will be used
```

@jgabry Dumb mistake on my end, just pushed an `any()` there which should take care of the warning

jgabry:
@avehtari Do you want to take a look and try this out before we merge it? 

@ecoronado92 In anticipation of merging this I just pushed a commit to this branch adding an item to the NEWS file and adding you as a contributor in authors list in the DESCRIPTION file.  I think this is pretty much ready to go, but let's also see what @avehtari says. 

In the meantime one more question about the speed. I just tried it on another example: 

```r
fit <- stan_glm(switch ~ I(dist/100) + log(arsenic), data = wells) # wells comes with rstanarm
y <- get_y(fit)
yrep <- posterior_predict(fit)
lw <- weights(psis(-log_lik(fit), r_eff = NA, cores = 4))
system.time(
  p <- ppc_loo_pit_overlay(y, yrep, lw, boundary_correction = TRUE)
)

   user  system elapsed 
226.708   1.501 228.761 
```

This dataset has about 3,000 observations and it took a 228 seconds to generate the plot. Is that consistent with the timings you were seeing or is that slower than expected?

ecoronado92:
```
This dataset has about 3,000 observations and it took a 228 seconds to generate the plot. Is that consistent with the timings you were seeing or is that slower than expected?
```
@jgabry hmm that does look slower than expected. 

 I ran the `radon` data set with 919 observations and 100 reference uniforms, it took 23 seconds. 
```
 user  system elapsed 
 22.337   0.223  23.286 
```

Then I just expanded it to have approx 3,000 obs using `rep` and that took
```
   user  system elapsed 
168.698   0.745 179.890 
```

The bottleneck was generating the reference lines but I'm wondering if plotting it might add extra time since we're not using `stat_density()` and instead rely on `geom_line()`.

avehtari:
I think this is too slow. I didn't expect that it could be this slow, but then I guess there is a reason why so many R packages include C/C++ code. 

Options: 1) find another faster package that makes that makes sensible KDE in closed range, 2) write own C/C++ code, 3) drop KDE density estimates and wait for a moment for ECDF based plots. What do you think?

ecoronado92:
```
Options: 1) find another faster package that makes that makes sensible KDE in closed range, 2) write own C/C++ code, 3) drop KDE density estimates and wait for a moment for ECDF based plots. What do you think?
```
@avehtari @jgabry  I can take a stab at option 2) this weekend using `Rcpp`. I'd probably use the Eigen library and would focus on speeding up the `bc_dunif` function (i.e. iterating through each element of the CDF value vectors). 

Then depending on the speed gain we can decide to continue via this path or wait until option 3) is available. Thoughts?

jgabry:
Thanks a lot for trying this out! That's indeed unfortunate that it doesn't make much of a difference. I took a brief look at the C++ code and I didn't see anything that jumped out at me as particularly inefficient (although I'm not a C++ expert), so this might just be a case where we can't get much of a speedup (possibly because `vapply()` is already pretty fast?). @avehtari Do you see anything in the C++ code that could be changed to speed it up?

R/bc_dunif.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+// Bias correction function (Chen 1999)
+double rho(double x,  double b) {
+  double c1 = 2.0*pow(b,2.0) + 2.5;
+  double c2 = sqrt(4.0*pow(b, 4.0) + 6.0*pow(b, 2.0) + 2.25 - pow(x,2.0) - x/b);
+  return c1 + c2;
+}
+
+// Helper function to generate dbeta value vector 
+VectorXd vec_dbeta(Map<VectorXd> x, double a, double b){
```
rok-cesnovar: Does this have to be passed in by value? I am also not sure on the reasoning behind using a map. `VectorXd& x` would pass by reference.

R/bc_dunif.cpp
```diff
@@ -0,0 +1,81 @@
...
+ double bc_kde_calc(double xs, double b,  NumericVector pvals){
+  
+  // Memory allocation
+  double d;
+  NumericVector out;
+
+  if ((xs >= 2*b) && (xs <= (1 - 2*b))){
+    out = Rcpp::dbeta(pvals, xs/b, (1 - xs)/b, 0.0);
+    Map<VectorXd>  tmp(Rcpp::as< Map<VectorXd> >(out));
+    d = tmp.sum() / tmp.size();               // vector mean
```
rok-cesnovar: I think this copy in line 40 could be avoided. 

as<Eigen::Map<Eigen::VectorXd> >(out).sum()/out.size() or something like that?

R/bc_dunif.cpp
```diff
@@ -0,0 +1,81 @@
...
+    d = tmp.sum() / tmp.size();
+  } else {
+    d = 0.0;
+  }
+  return d;
+}
+
+
+// [[Rcpp::export]]
+RowVectorXd bc_dunif(Map<VectorXd> xs, NumericVector pvals,
```
rok-cesnovar: Same as above: VectorXd& xs. 

Does NumericVector have to be passed by value? This might be a rcpp thing.

rok-cesnovar (commented):
Disclaimer: While I know my way around c++, I have less experience with RCpp and the limitations/rules it has, so these comments might be useless.

I am posting a couple of suggestions.

jgabry:
Thanks @rok-cesnovar!

R/bc_dunif.cpp
ecoronado92: @rok-cesnovar Thanks for going over the code! I think I forgot to remove this function, this isn't currently used in the code. I hadn't been able to get `Rcpp::dbeta` (inherently works with vecs) to work so was using `R::dbeta` with this as a helper function.

R/bc_dunif.cpp
ecoronado92: Good point! I'll change it

R/bc_dunif.cpp
ecoronado92: Yep, same as previous comment.

ecoronado92:
@jgabry I'll wait to hear from @avehtari in case he has any more suggestions and work on it during the weekend. Currently started a new job and am a bit swamped with work but would love to continue contributing to this feature.

In the meantime I'll make the changes as @rok-cesnovar suggested (big thanks!) and will see if that helps at all. I guess at the for loop level the only way to speed things is running these in parallel.

avehtari:
I don't see anything obvious. I guess copies may have a significant performance hit, but it might be also all the dbeta's. How many dbeta's need to be computed?

R/ppc-loo.R
```diff
@@ -456,3 +519,81 @@ ppc_loo_ribbon <-
...
+  # Re-scaling inputs based on upper_bound given beta kernel is defined [0,1]
+  xs <- xs/xmax
+  pvals <- pvals/xmax
+  b <-  b/xmax # smoothing parameter (i.e. bw)
+  
+  d <- vapply(X = xs, 
+              FUN = .bc_kde_calc, 
+              b = b,
+              pvals = pvals, 
+              FUN.VALUE = 0)
```
hhau, line 539: Hello! I don't mean to chime in out of nowhere with a potentially useless suggestion, but one can usually get a reasonable performance bump by writing an `Rcpp` function that takes a `NumericVector` or `NumericMatrix` and does the looping / `vapply`ing inside `C++` instead of in `R` (at least in my own experience translating KDEs to `Rcpp`). 

It may be worth looking at something like [`sm::density`](https://github.com/cran/sm/blob/master/R/density.r), not for code (it's all in R) but for the methodology (make everything a vector/matrix and rely on the linear algebra operations as much as possible), as it is surprisingly speedy (sometimes) for a pure R implementation as a result.

ecoronado92:
@avehtari sorry for the late reply. 

>How many dbeta's need to be computed?

One per each element of each simulated `runif()`. So if we're simulating 100 samples of length 3000, then 100*3000 = 300k `dbetas` are being computed. The `pvals` in the `dbeta` don't change, it is the `xs` (elements of a `runif()` ) that change.

Workarounds are:
1) parallelize with current setup
2) find a way to apply the piecewise function (i.e. set of if-else in `bc_dunif()`) in one go without iteration

avehtari:
>  So if we're simulating 100 samples of length 3000, then 100*3000 = 300k dbetas are being computed. The pvals in the dbeta don't change.

But if there are several pvals, then it's 300k * number of pvals dbeta calls?

avehtari:
Arviz uses reflection trick with good looking results. See plot_loo_pit in ArviZ https://arviz-devs.github.io/arviz/generated/arviz.plot_loo_pit.html#arviz.plot_loo_pit

ecoronado92:
@avehtari @jgabry Got some promising news! Just implemented the reflection trick in a quick code snipped and we have something that might just work.

Here's an example of the output (grey lines are reference, blue is pit values):

Screen Shot 2020-08-29 at 3 11 57 PM

Here's an example of how long it took to run on the radon example with `n=919` samples and generating the 100 reference uniforms

```
   user  system elapsed 
 10.711   0.646  11.847 
```

Next step is to try it with @jgabry's 3k observation example and see what's the speed. Hopefully I have something tangible tomorrow.

avehtari:
Great!

ecoronado92:
@avehtari I ran it with @jgabry example and this is what came out (but I'm unsure the plot is correct). I tried it also with other examples (radon data, some simulated data I had) I had tested and these seemed to look fine so maybe this is what the plot should look like.

Screen Shot 2020-08-30 at 5 35 54 PM

In terms of speed, we are in good territory
```
  user  system elapsed 
  0.348   0.009   0.412 
```

I'll be pushing these new code for you both to play around if you'd like.

avehtari:
```
fit <- stan_glm(switch ~ I(dist/100) + log(arsenic), data = wells ...
```
Switch is binary and discrete data should be handled differently. See Predictive Model Assessment for Count Data https://doi.org/10.1111/j.1541-0420.2009.01191.x

ecoronado92:
> Switch is binary and discrete data should be handled differently

@avehtari  Thanks! Will have to take a look this weekend.

jgabry:
Just realized I forgot to comment on this. Glad to see that trick speeds it up a lot! Thanks a lot! Let me know when this is ready for another code review.

ecoronado92:
@jgabry Sorry I haven't been able to get to this, just started a new job and I've been a bit swamped over the weekdays. I'll take a look at it this weekend + @avehtari thanks for reaching out to Tomas! He shared with me some code and other KDE implementations he's done which I think will be insightful to get this finalized.

jgabry:
> Sorry I haven't been able to get to this, just started a new job and I've been a bit swamped over the weekdays.

Congrats on the new job. No rush!

ecoronado92:
@avehtari I read the paper and had a quick question just to make sure I implement the right thing.

My understanding is that the PIT values are generated via the non-randomized piecewise function to make them comparable with a normal. However, I'm confused whether these non-randomized values are just PIT values that get transformed from previous ones, or if they are generated from scratch - count/binary data.

So my question is, should I transform the PSIS-LOO generated values in `loo()`, or do you think we need new feature to handle these via PSIS-LOO?

If it's the former I think I can do it fairly quickly, otherwise it might take some time given I haven't delved into that code.

avehtari:
Binary data is a bit strange for PIT, so for binary data I would suggest plots like in https://avehtari.github.io/modelselection/diabetes.html#44_calibration_of_predictions

For other discrete data, you can also start with the randomized version as it's easier to understand and implement.
Maybe [this discourse post](https://discourse.mc-stan.org/t/hierarchical-multivariate-ordinal-model-diagnostics-and-tuning/16933/6) helps

ecoronado92:
@avehtari I see, thank you!

So just so I understand, here's what I understand needs to be done (@jgabry keep me honest):

1) If data is binary, add a check so the following plots are generated instead of the KDE plots at the moment
>Binary data is a bit strange for PIT, so for binary data I would suggest plots like in https://avehtari.github.io/modelselection/diabetes.html#44_calibration_of_predictions

2) Implement a randomized version for discrete data, again a check in the code
Could you clarify what N, p, n should be in our case?
>n<-1000000;N<-100;p<-.3;r<-rbinom(n, N, p);

avehtari:
> If data is binary, add a check so the following plots are generated instead of the KDE plots at the moment

Or check and give a warning that this plot for binary is not that useful and other plotting function would be better.

> Implement a randomized version for discrete data, again a check in the code
Could you clarify what N, p, n should be in our case?

Those values were for a simulated data, and we would use the values that are in the data.

> Otherwise, can I suggest we merge this for continuous data and keep working on the other data types on a separate issue?

Just add to the documentation the mention that the current one is for continuous valued observations.

aloctavodia:
I think you may be interested in arviz-devs/arviz#1379. We have been experimenting with interpolation for discrete variables, it seems to work ok.

ecoronado92:
@jgabry I have pushed the warnings for binary data and think this should be good to review again!

> Implement a randomized version for discrete data, again a check in the code
Could you clarify what N, p, n should be in our case?

>Those values were for a simulated data, and we would use the values that are in the data.

@avehtari So in our case, let's say I get `dim(yrep)` being a 4000x3020 matrix using some count dataset. Would a pseudo-code look like this?

```
N <- ncol(yrep) #3020

for each row i in yrep
     r <- row_yrep_i
     p <- mean(r >0.5)
     n <- 100
     rand_pit_i <- pbinom(r-1, N, p)+runif(n)*(pbinom(r, N, p)-pbinom(r-1, N, p))

pit_mat[i]i<- rand_pit_i # a 4000x3020 matrix of randomized pit values
```

avehtari:
@aloctavodia 
> I think you may be interested in arviz-devs/arviz#1379. We have been experimenting with interpolation for discrete variables, it seems to work ok.

Is this the non-randomized PIT as in https://doi.org/10.1111/j.1541-0420.2009.01191.x ?
I looked at that PR, but the discussion was not clear on what was actually implemented (or I missed that)

avehtari:
> Would a pseudo-code look like this?

I'll have time to check the pseudocode (or provide new pseudocode) in the beginning of October.

aloctavodia:
@avehtari 

I still have to read that paper.  The main changes are lines https://github.com/arviz-devs/arviz/blob/master/arviz/plots/backends/matplotlib/bpvplot.py#L91-L98 i.e. for discrete variables `obs_vals` and `pp_vals` are replaced by interpolated data.

jgabry:
> @jgabry I have pushed the warnings for binary data and think this should be good to review again!

Thanks, looks great!

jgabry:
@avehtari Now that the boundary correction is much faster with the reflection trick, what do you think about defaulting to `boundary_correction=TRUE`?

avehtari:
defaulting to `boundary_correction=TRUE` would be good

ecoronado92:
@jgabry I just changed boundary correction to TRUE in functions and all documentation. I also made sure to pull from master before adding this commit to make sure it's up to date with any changes there.

jgabry:
> @jgabry I just changed boundary correction to TRUE in functions and all documentation. I also made sure to pull from master before adding this commit to make sure it's up to date with any changes there.

Thanks! I think this is ready to go, just one last thing that I hadn't noticed before (sorry!). Right now there's a comment in the code

```
  # TODO currently boundary correction generates NAs for values near ends of vector
  # after convolution . Important to note that # of NAs "varies" so na.omit()
  # won't work and will cause size inconsistencies
  # Current fix, use last non-null value at head and tail of vector as padding
```

Is there anything that still needs to be done here? If so, can you move it from a code comment to a new issue? We can merge this with the current fix, but we'd prefer to avoid using code comments to keep track of TODO items.  

Otherwise I think that's it. Thanks again for working on this, it's a big improvement!!

ecoronado92:
> Is there anything that still needs to be done here? If so, can you move it from a code comment to a new issue? We can merge this with the current fix, but we'd prefer to avoid using code comments to keep track of TODO items.

@jgabry Yes, I wasn’t able to figure why this was happening so I’m happy to move this to a separate issue. I’ll make changes later today and will ping you

jgabry:
Awesome, thanks!

ecoronado92:
@jgabry Removed the TODO from the code and created a new issue pointing at where the solution should be addressed in the code.

Let me know if we need to do anything else :)

avehtari:
@ecoronado92 It would be useful to mention in the help text that the boundary correction is using the "reflection method" (method is better word than trick here). Even better if there would also be an article reference. Did ArviZ have a reference?

@jgabry would you have time to re-create the bayes-vis paper LOO-PIT figures with the new version? If you don't have time I'll do it.

ecoronado92:
@avehtari Just added reflection method to the documentation and code comments.

>Did ArviZ have a reference?

Here's the only references I found related to their LOO PIT function and/or the reflection method. 
```
References
    ----------
    * Gabry et al. (2017) see https://arxiv.org/abs/1709.01449
    * https://mc-stan.org/bayesplot/reference/PPC-loo.html
    * Gelman et al. BDA (2014) Section 6.3
```

jgabry:
> Even better if there would also be an article reference. Did ArviZ have a reference?

Yeah I agree it would be good to cite the reflection method but I don't see a reference for that in arviz either. Just the reference to the preprint of our paper, like @ecoronado92 found. 


> @jgabry would you have time to re-create the bayes-vis paper LOO-PIT figures with the new version? If you don't have time I'll do it.

Yeah I have a few min right now so I will try to do that quickly.

jgabry:
@avehtari Ok here are PIT plots from our paper using the new version: 

ppc_loo_pit_overlay-new-4

One thing I notice is that we still have the axis limits cut off before 0 and 1, but now that we have the boundary correction we could let the axis limits include 0 and 1. What do you think? This would need to get changed in the ppc_loo_pit_overlay() source code but it's a one line change. 

Also, just as a reference, here are the same plots with `boundary_correction=FALSE`: 

ppc_loo_pit_overlay-new-5

avehtari:
I asked in ArviZ slack for a reference

Thanks for the plots. They look better, and I think the interval should from 0 to 1.

avehtari:
@tomicapretto and @aloctavodia helped to find a useful reference:
Boneva, L. I., Kendall, D., & Stefanov, I. (1971). Spline transformations: Three new diagnostic aids for the statistical data-analyst. Journal of the Royal Statistical Society. Series B (Methodological), 33(1), 1-71.
https://www.jstor.org/stable/2986005
See figs 5 and 6
image

jgabry:
Ok great, I just pushed a commit citing that Boneva et al paper and I made the x-axis limits go out to 0 and 1 if boundary_correction=TRUE. Here's what the plots look like now: 

ppc_loo_pit_overlay-new-6

@avehtari Anything else before we merge this?

avehtari (approved):
Looks good.

jgabry (approved):

jgabry:
Ok I think we're ready to merge this. Thanks again for implementing this!

PR #243, opened 2020-10-20
Fix y axis for ppc_loo_pit_qq() with 'compare = "normal"'

### LINKED ISSUE #240: ppc_loo_pit_qq() with 'compare = "normal"'
Since `ppc_loo_pit_qq()` with `compare = "normal"` adds a straight line, I wonder if the line
https://github.com/stan-dev/bayesplot/blob/5500d2eecb6beff9ab1654d0a79bc9b591fad270/R/ppc-loo.R#L223
shouldn't better read
```r
pit <- as.vector(stats::qnorm(pit))
```
(and the documentation be adjusted accordingly).

fweber144:
This fixes #240. Notes:
* I have updated the documentation as well as possible. But I might have missed some places in the documentation which need to get updated, too.
* I have moved two lines of code in "tests/testthat/test-ppc-loo.R" which I think got misplaced.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -10,7 +10,7 @@
 
 * On the y axis, `ppc_loo_pit_qq(..., compare = "normal")` now plots standard
   normal quantiles calculated from the PIT values (instead of the standardized
-  PIT values). (#240, #<PR_number>, @fweber144)
+  PIT values). (#240, #243, @fweber144)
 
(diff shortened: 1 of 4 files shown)


jgabry (approved):
Looks great, thanks! And thanks for adding the NEWS item.

PR #246, opened 2020-10-29
Use melt_mcmc instead of calling reshape2::melt

### LINKED ISSUE #245: mcmc_acf errors for some input types
Reprex (using the github version):

``` r
library(bayesplot)
#> This is bayesplot version 1.7.2.9000

mcmc_acf(
  array(
    data = rnorm(100 * 2 * 2), 
    dim = c(100, 2, 2),
    dimnames = list(
      Iteration = 1 : 100,
      Chain = 1 : 2,
      Parameter = 1 : 2
    )
  )
)
#> Error in factor(rep(1:n_param, each = n_chain * n_lags), labels = levels(data[["Parameter"]])): invalid 'labels'; length 0 should be 1 or 2
```

<sup>Created on 2020-10-29 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

I think this is because here: https://github.com/stan-dev/bayesplot/blob/master/R/mcmc-diagnostics.R#L583 you directly call `reshape2::melt`, but you mean to call https://github.com/stan-dev/bayesplot/blob/master/R/helpers-mcmc.R#L125 (because I think you've run into this issue before somehow 🙂).

hhau:
Closes #245 (tests pass locally for me)

### DIFF
--- a/R/mcmc-diagnostics.R
+++ b/R/mcmc-diagnostics.R
@@ -580,8 +580,7 @@ acf_data <- function(x, lags) {
     abort(paste0("Too few iterations for lags=", lags, "."))
   }
 
-  data <- reshape2::melt(x, value.name = "Value")
-  data$Chain <- factor(data$Chain)
+  data <- melt_mcmc(x)
   ac_list <- tapply(
     data[["Value"]],
     # INDEX = list(data[["Chain"]], data[["Parameter"]]),
--- a/NEWS.md
+++ b/NEWS.md
@@ -32,6 +32,7 @@
   can be set to FALSE to recover the old version of the plot. (#171, #235,
   @ecoronado92)
 
+* Fix error in `mcmc_acf()` for some input types. (#244, #245, @hhau)
 
 # bayesplot 1.7.2
 


R/mcmc-diagnostics.R
```diff
@@ -580,8 +580,7 @@ acf_data <- function(x, lags) {
     abort(paste0("Too few iterations for lags=", lags, "."))
   }
 
-  data <- reshape2::melt(x, value.name = "Value")
-  data$Chain <- factor(data$Chain)
```
jgabry, line 584: Is there a reason to also get rid of making `Chain` a factor? To be honest I'm not sure why that was there in the first place (it may never have been necessary, maybe I thought `tapply` required factors), but curious if you removed that intentionally?

jgabry (commented):
Good catch, thanks for this! I just put one tiny question below (or above, wherever it ends up).

R/mcmc-diagnostics.R
hhau: I removed it intentionally. I didn’t think it was necessary as `tapply` calls `as.factor` on `INDEX` anyway.

jgabry: Ok that makes sense, thanks!

jgabry (approved):
Thanks again for the PR. Merging now.

Issue #242, opened 2020-10-20
mcmc_intervals: add example of relabeling without changing order


jscamac:
Hi, I'm attempting to relabel the parameters in a `mcmc_interval` plot by adding a `ggplot2::scale_y_discrete(labels = new_labels)` approach.

The problem I'm having is that when I do this, the labels are correctly relabelled, but the order of the yaxis changes, such that the Intercept term now is at the bottom as opposed to the top (I prefer to have the intercept on top). I've played around with trying to set the breaks or inverting the limits of the axis but have had no success. Also because the mcmc plot functions make a variety of changes to the underlying data prior to plotting I don't really see an obvious way to set the order. Is there something obvious I'm missing here?

```
# MWE example
prior1 <- prior(normal(0,10), class = b) +
  prior(cauchy(0,2), class = sd)
fit1 <- brm(count ~ zAge + zBase * Trt + (1|patient),
            data = epilepsy, family = poisson(), prior = prior1)

# Default (no relabelling but correct order)
mcmc_intervals(fit1, regex_pars="b_"

# Relabelling (but order of pars changes)
mcmc_intervals(fit1, regex_pars ="b_") +
  ggplot2::scale_y_discrete(labels = c("b_Intercept" ="Intercept",
                                       "b_zAge" = "Age",
                                       "b_zBase" = "Base",
                                       "b_Trt1" = "Treatment",
                                       "b_zBase:Trt1" = "Base x Treatment"))
```

jscamac:
Nevermind... I was obviously missing something:

The solution is:

```
mcmc_intervals(fit1, regex_pars ="b_") +
                 ggplot2::scale_y_discrete(labels = c("b_Intercept" ="Intercept",
                                                      "b_zAge" = "Age",
                                                      "b_zBase" = "Base",
                                                      "b_Trt1" = "Treatment",
                                                      "b_zBase:Trt1" = "Base x Treatment"),
                                           limits = c("b_zBase:Trt1",
                                                      "b_Trt1",
                                                      "b_zBase",
                                                      "b_zAge",
                                                      "b_Intercept"))
```

jgabry:
Glad you found the solution! Just seeing this now but you ended up finding exactly what I would have a recommended. Is there something we could put in the doc that would have made it easier to figure this out? Or maybe we should just add an example of doing this to the Examples section in the doc?

jscamac:
[...]

PR #247, opened 2020-11-03
adding 2 examples of relabeling without changing order

cbemben:
Added 2 examples to demonstrate the ordering component. Happy to limit it to 1 example if that's cleaner. #242


jgabry (approved):
@cbemben Thanks for the PR! This looks great and all the checks are passing so merging now.

PR #254, opened 2020-12-09
Fix CRAN issues

jgabry:
I'm opening this PR so it triggers running R cmd check on R-devel so I can look at the results. Apparently there are test failures to fix but only on R-devel. Eventually I'll use this PR to make the necessary changes.


jgabry:
Looks like R cmd check on windows is failing because of the issue that this RStan PR fixed https://github.com/stan-dev/rstan/pull/857. That PR was merged but hasn't made it into a release yet. Otherwise everything is passing now, even on R devel.

Issue #256, opened 2021-01-04
ppc ribbon with points for observations


charlesm93:
Per the discussion on https://discourse.mc-stan.org/t/bayesplot-ppc-ribbon-using-points-for-observations/19599, it'd be nice to have a ppc plot which uses a ribbon for the model predictions but points for the observations.

I propose adding an argument to the function `ppc_ribbon` and `ppc_ribbon_grouped`, namely `y_is_point` which defaults to FALSE (so as to not break backward compatibility). So the signature would be

```
ppc_ribbon(
  y,
  yrep,
  x = NULL,
  ...,
  prob = 0.5,
  prob_outer = 0.9,
  alpha = 0.33,
  size = 0.25,
  y_is_point = FALSE
)
```

I can build this feature in.

jgabry:
Thanks Charles, I'm in favor of this. Regarding the argument, in the `ppc_violin_grouped()` function we currently have an argument `y_draw` with three options `y_draw="violin"`,  `y_draw="points"` or `y_draw="both"`. Could we just do the same here but the first option is `"line"` instead of `"violin"`?

charlesm93:
Sure thing. I'll have `y_draw` default to "line".

jgabry:
closing this since we merged the PR already

PR #257, opened 2021-01-08
Feature ppc ribbon obspoints

charlesm93:
Addresses issue #256 

Add an additional argument, `y_draw`, for `ppc_ribbon` and `ppc_ribbon_grouped` which specifies whether the observed y should be plotted using a `point`, `line`, or `both`.

While at it, add unit tests for `ppc_ribbon_grouped`, emulating what's be done for `ppc_intervals_grouped`.


charlesm93:
@jgabry can you review? Thank you!

charlesm93:
This PR may have been a bit premature. What's the documentation policy? Is a standard manual that needs to be updated? What about the vignettes, should some get added?

I'm not sure why the tests fail. It seems to be related to installing dependencies.

jgabry:
Hey Charles, really sorry for the delay on this. I just pushed a few commits fixing a few things, but this looks good. I'll add some review comments pointing out the changes.

R/ppc-intervals.R
```diff
@@ -173,14 +173,17 @@ ppc_intervals_grouped <- function(y,
 
 #' @rdname PPC-intervals
 #' @export
+#' @param y_draw For ribbon plots only, a string specifying how to draw `y. Can
+#'   be `"line"` (the default), `"points"`, or `"both"`.
```
jgabry: One reason CI was failing was because of a missing doc entry for the y_draw argument. I added that here.

R/ppc-intervals.R
```diff
@@ -173,14 +173,17 @@ ppc_intervals_grouped <- function(y,
...
 ppc_ribbon <- function(y,
                        yrep,
                        x = NULL,
                        ...,
                        prob = 0.5,
                        prob_outer = 0.9,
                        alpha = 0.33,
-                       size = 0.25) {
+                       size = 0.25,
+                       y_draw = c("line", "points", "both")) {
```
jgabry, line 186: Not essential, but it's common in R to put all valid options in a vector and then use `match.arg` later to ensure that what the user wrote is a valid option. (See below where I added `match.arg`)

R/ppc-intervals.R
```diff
@@ -319,9 +325,11 @@ label_x <- function(x) {
            size = 1,
            grouped = FALSE,
            style = c("intervals", "ribbon"),
-           x_lab = NULL) {
+           x_lab = NULL,
+           y_draw = c("line", "points", "both")) {
 
   style <- match.arg(style)
+  y_draw <- match.arg(y_draw)
```
jgabry, line 332: I added this because the `match.arg` function will result in an error if the user's `y_draw` argument isn't one of the three valid ones.

jgabry (approved):
With the small changes I pushed let's see if all the checks pass now. But this looks good to me!

R/ppc-intervals.R
charlesm93: Makes sense. I was wondering what match.arg does.

charlesm93:
@jgabry All the changes you made look good to me. Looks like we still have one failing tests, concerning the vignettes. I don't believe I made any changes to those, but if something needs fixing, I can get to it.

jgabry:
The one failing check is unrelated to this PR, so I'm going to go ahead and merge. Sorry for the delay!

PR #260, opened 2021-02-01
New function ppc_km_overlay_grouped()

fweber144:
I added `ppc_km_overlay_grouped()`. Unfortunately, the `+ list(data)` trick which is used e.g. in `ppc_ecdf_overlay_grouped()` cannot be used in `ppc_km_overlay_grouped()` since if it was used, `survival::survfit()` would not "know" of the grouping and would not be able to take it into account. The solution I have implemented now is kind of hacky. In particular, it requires to pass the grouping to `ppc_km_overlay()` which I did via the ellipsis (`...`). I couldn't think of a better solution, but if you have an idea, please let me know.

I also performed some minor fixes, e.g. correcting a typo in "args-groups.R" and updating the expected results for the "cmdstanr methods work" unit test. Just have a look at my commits and decide which ones you want to use.


fweber144:
The failing check on Windows might arise from [this "withr" issue](https://discourse.mc-stan.org/t/new-error-cleanup-makevar-old-argument-rmu-is-missing-with-no-default/18137).

fweber144:
No problem, you must have a lot of work with all those R packages. Thanks for merging!

PR #265, opened 2021-03-15
fixes #264 (wrong order of the parameters in mcmc_recover_hist)

### LINKED ISSUE #264: mcmc_recover_hist with arrays shows the wrong order of the facets
`mcmc_recover_hist` messes the order of the parameters.

Reprex below:
```r
m <- '
data {
}
parameters {
  real<lower=0,upper=1> theta[25];
}
model {
  theta ~ beta(1,1);  // uniform prior on interval 0,1
}'

mstan <- cmdstan_model(write_stan_file(m))

fit <- mstan$sample(
  data = list(),
  seed = 123,
  parallel_chains = 4
)
```
This is fine:

```
mcmc_hist(fit$draws("theta"))
```

But here the order of the thetas is alphanumeric, rather than the order of the parameters
```
mcmc_recover_hist(fit$draws("theta"), rep(.5, 25))
```

version:
`bayesplot_1.8.0 `

bnicenboim:
fixes https://github.com/stan-dev/bayesplot/issues/264


mike-lawrence:
Think this would also fix the similar wrong-order of `mcmc_recover_intervals()`?

bnicenboim:
by the way, the fail in windows is an unrelated problem in vignette (it seems)

bnicenboim:
any news with this?

jgabry:
Really sorry for the delay on this! Thanks for the fix. I agree that the windows failure is unrelated to this. Will merge now.

PR #269, opened 2021-05-24
Require ggfortify for some tests

billdenney:
While working on a different PR, I noticed that I do not have ggfortify installed and tests failed because of that.  This PR skips those tests when ggfortify is not installed.


jgabry (approved):
Thanks!

PR #270, opened 2021-05-24
Make mcmc_pairs() return a ggplot object instead of a bayesplot_group object (Fix #268)

### LINKED ISSUE #268: Feature Request: Make bayesplot_grid objects a ggplot object
Related to #128 and to several other questions (such as https://discourse.mc-stan.org/t/manipulating-bayesplot-plots/4419).

It would be handy if `bayesplot_grid` objects could be modified like any other ggplot object.  I understand that it is not currently a ggplot object is "It wasn't possible (at least it didn't seem possible when we tried) to get the plot how we wanted it to look without creating separate plots and combining them in a grid like that." (according to @jgabry in https://github.com/stan-dev/bayesplot/issues/128#issuecomment-345466831).

But, as I am running into issues much like those in #128 where I want to modify the figure after it is created, it would be helpful to use normal ggplot mechanics to do so.

I'm trying to work around this by using `cowplot::plot_grid()` or similar functions on the `return_value$bayesplot` list of plots.

Is there interest in converting a bayesplot_grid object into something that can be used with normal ggplot2?  (If so, I can try to generalize what I'm working on into a PR.  If not, I won't.)

billdenney:
This PR makes `mcmc_pairs()` return a ggplot object instead of a bayesplot_group object (Fix #268).

This is a DRAFT PR for review and comment!!!  The reason that it is a draft is that it emits all sorts of warnings that would need to be fixed before merging, but I know that there are strong preferences about the graphical organization of the output of `mcmc_pairs()`, and I wanted to provide this draft to allow for testing and review.  For a final PR, I would ensure that warnings were not present (or that they were minimized as much as possible).

I believe that I captured all the major functionality from the existing `mcmc_pairs()` output, but there were some interactions around how divergent transitions, max tree depth hits, and hex plots that may be off.


billdenney:
Some differences that I think are notable are:

1. x-axes and y-axes are aligned now (because they are facets)
2. row and column titles are now in strip labels instead of subtitles of the diagonal elements (again because they are facets)
3. the vertical scale for the density or histogram plots along the diagonal are the full height of the facet rather than sharing the height with the subtitle

jgabry:
Hey @billdenney thank you for this and sorry for the slow reply! I'm caught up in a bunch of other stuff at the moment but hopefully I'll have time to give you some comments soon.

PR #271, opened 2021-05-27
Add Plot to Show Variation in LOO-CV

ParadaCarleton:
Adds the plot requested [here](https://github.com/stan-dev/loo/issues/127).

### DIFF
--- a/R/model-comparison-stacking.R
+++ b/R/model-comparison-stacking.R
@@ -1,0 +1,155 @@
+#' Find opportunities for model stacking 
+#' 
+#' The pancake plot shows how the predictive accuracy of two different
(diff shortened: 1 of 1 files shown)


ParadaCarleton:
Adds the plot requested [here](https://github.com/stan-dev/loo/issues/127).

ParadaCarleton:
This is mostly done, but I would like to add one more example showing the plot used together with model stacking. I'm wondering where I can pull a dataset that shows model A outperforming when some continuous variable `x` is small, but model B outperforming when it's large. I'd like to compare this behavior to the stacked model doing well for all values of `x`.

ParadaCarleton:
Moved to LOO package

Issue #267, opened 2021-05-06
Add ECDF plot and ECDF difference plot with simultaneus confidence intervals as graphical goodness-of-fit and convergence diagnostics


TeemuSailynoja:
As requested by @avehtari and @hyunjimoon, implement the ecdf plots from [this article](https://arxiv.org/abs/2103.10522). Code for the article is available [here](https://github.com/TeemuSailynoja/simultaneous-confidence-bands/blob/main/ecdf_plot_functions.R), so mostly cleaning and formatting needed.

jgabry:
This would be great!

hyunjimoon:
Wow thanks, Teemu! I have seen different versions of binomial/beta confidence intervals so having this included in bayesplot would be a great help to users being less confused about SBC overall.

TeemuSailynoja:
Closed by #282

PR #273, opened 2021-07-15
issue 267 - ECDF plots 

TeemuSailynoja:
The first implementation of the ecdf plots mentioned in #267 . Currently, plots the mentioned plots for:
- precomputed PIT values (one sample and multiple sample comparison)
- Empirical PIT values when `y` and `yrep` are provided. Currently assumes `y`to consist of multiple draws from a univariate distribution -> PIT for multivariate `y` to be implemented later.
- Fractional ranks for multiple sample comparison when only `yrep` is provided. Here each chain is ranked in relation to the total sample obtained by merging all of the chains together. 

I decided to add the functions into `ppc-intervals.R` as the share most similarities with the functions in that file.


R/helpers-ppc.R
```diff
@@ -57,14 +58,35 @@ validate_yrep <- function(yrep, y) {
...
+#' Validate PIT
+#'
+#' Checks that the provided probability integral transformation (PIT) values are
+#' a numeric matrix with no NAs and that the provided values fall in [0,1].
+#'
+#' @param pit The 'pit' object provided by the user.
+#' @return Either throws an error or returns a numeric matrix.
+#' @noRd
+validate_pit <- function(pit) {
+  stopifnot(is.matrix(pit), is.numeric(pit))
```
avehtari, line 78: Why pit needs to be a matrix? The ecdf plot function would be useful also for vectors.

R/helpers-ppc.R
```diff
@@ -250,6 +272,157 @@ all_counts <- function(x, ...) {
...
+  }
+  lims$lower <- c(rep(lims$lower[1:K], each=2), lims$lower[K + 1])
+  lims$upper <- c(rep(lims$upper[1:K], each=2), lims$upper[K + 1])
+  lims
+}
+
+#' Transform observations in 'x' into their corresponding fractional ranks.
+u_scale <- function(x) {
+  array(rank(x) / length(x), dim = dim(x), dimnames = dimnames(x))
+}
```
avehtari, line 418: Why this is defined differently from the posterior package?

avehtari:
If I provide gamma different from what the code computes, I get two set of bands
image

TeemuSailynoja:
> First
> 
> ```
> empirical_pit <- function(y, x) {
>      N = length(x)
>      (colSums(outer(x, y, "<=")) + 1) / (N + 1)
>  }
> y=rnorm(N);x=rnorm(N);q=empirical_pit(y, x);
> ```
> 
> then this works
> 
> ```
> > ppc_ecdf_intervals(pit=matrix(q,nrow=1),gamma=gammax)
> ```
> 
> but this doesn't
> 
> ```
> > ppc_ecdf_intervals_difference(pit=matrix(q,nrow=1),gamma=gammax)
>  Error: Aesthetics must be either length 1 or the same as the data (1001): y, colour and group
> Run `rlang::last_error()` to see where the error occurred. 
> ```

The aesthetics of ppc_ecdf_intervals_difference were misaligned due to a misplaced parenthesis. This is now fixed. Thank you for pointing this out.

TeemuSailynoja:
> If I provide gamma different from what the code computes, I get two set of bands image

I chose both `ppc_ecdf_intervals` and `ppc_ecdf_intervals_difference` to draw by default confidence bands for both 50% and 90%, as these are the default values for the other `ppc_intervals` functions.
If this is confusing and only one set of confidence intervals would be preferrable, I can disable the other pair by default.

R/helpers-ppc.R
TeemuSailynoja: This should be extended and tested for vectors.

R/helpers-ppc.R
TeemuSailynoja: I had not seen the counterpart in the posterior package. There the fractional ranks seem to be normalized to { 1 / (2S), ..., 1 - 1 / (2S) }, where S is the number of values in the combined sample. Is the idea there to move the fractional ranks to the middle of the respective "bins" instead of having them at the "edges" as is implemented above?

avehtari: It would be good to use the same function or use different name for the function if the functionaility is different.
The posterior::u_scale sets the values so that transformation to normal would produce values that are close to normal ordered statistics (while 0 and 1 would transform to -Inf and Inf, and the rule now used is the best according to Blom)

avehtari:
> I chose both ppc_ecdf_intervals and ppc_ecdf_intervals_difference to draw by default confidence bands for both 50% and 90%, as these are the default values for the other ppc_intervals functions.
> If this is confusing and only one set of confidence intervals would be preferrable, I can disable the other pair by default.

It is confusing that setting gamma, changes only one of the bands and I get lines that are often overlapping, and it's confusing that the color and thickness of the line are the same

TeemuSailynoja:
Reworked in #282.

PR #222, opened 2020-06-16
Add new PPD module for plots without `y` variable

jgabry:
Closes #151 
Advances #97
Advances #196

This is a big PR (sorry!) that adds a lot of new functions and refactors a bunch of existing code along the way. The bad news is that there are a ton of changed files, the good news is that many of them are generated .Rd files and svgs for vdiffr, although still many files with changed R code. @tjmahr Given how large this PR is, what would be the way that I could best help you go through this? For example, do you want me to generate and post images of all the new plots? Or schedule a time to talk through some of it? Anyway, totally up to you, just let me know what I can do to help. 

The major new addition is a new "module" PPD -- see `help("PPD-overview")` -- which is for plotting posterior or prior predictive distributions without comparing to any observed data. These functions have prefix `ppd_` instead of `ppc_` and take an argument `ypred` instead of `yrep` and `y`. 

Some additional refactoring (discussed in https://github.com/stan-dev/bayesplot/issues/151#issuecomment-496323557 and subsequent comments) makes it so that `_grouped()` functions no longer require duplicating much of the same code from the non-grouped function. Instead, the non-grouped function is called and then faceting is added. 

#### New PPD functions

PPD-distributions
  - ppd_data()
  - ppd_dens_overlay()
  - ppd_ecdf_overlay()
  - ppd_dens()
  - ppd_hist()
  - ppd_freqpoly()
  - ppd_freqpoly_grouped()
  - ppd_boxplot()

PPD-intervals
  - ppd_intervals_data()
  - ppd_intervals()
  - ppd_intervals_grouped()
  - ppd_ribbon()
  - ppd_ribbon_grouped()

PPD-test-statistics
  - ppd_stat_data()
  - ppd_stat()
  - ppd_stat_grouped()
  - ppd_stat_freqpoly()
  - ppd_stat_freqpoly_grouped()
  - ppd_stat_2d()

#### More ppc_*_data() functions

Also, it made sense while refactoring to add the following `ppc_*_data()` functions that were missing: 

  - ppc_bars_data()
  - ppc_error_binnned_data()
  - ppc_error_data()
  - ppc_scatter_avg_data()
  - ppc_scatter_data()
  - ppc_stat_data()

#### More functions get facet_args arguments that were missing

  - ppc_scatter()
  - ppc_scatter_avg_grouped()
   - ppc_error_hist()
   - ppc_error_hist_grouped()
   - ppc_error_scatter()
   - ppc_error_binned()

PR #277, opened 2021-10-06
handle posterior's draws objects

### LINKED ISSUE #274: Add support for rvars draws format
To avoid need for manually switching between different draws formats, it would be nice if bayesplot functions would allow draws_rvars format as input. At least `mcmc_hist` already works with draws_array, draws_matrix and draws_df.

Ozan147:
Closes #274 

Provides the flexibility to use all [posterior](https://github.com/stan-dev/posterior) `draws` objects as inputs to plotting functions by casting them to `draws_array` within `prepare_mcmc_array()`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 inst/doc
 
 tests/testthat/Rplots.pdf
+tests/testthat/_snaps/
 
 .DS_Store
 
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -31,6 +31,7 @@ Imports:
     ggplot2 (>= 3.0.0),
     ggridges,
     glue,
+    posterior,
     reshape2,
     rlang (>= 0.3.0),
     stats,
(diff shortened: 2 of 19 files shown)


jgabry:
Thanks Ozan! The code changes look good. I'd like to test it out before merging but I have a bit of a full day today. I should be able to try it out later today or tomorrow at the latest. 

Also, don't worry about the failing checks on windows and mac for now. I think there's an issue unrelated to this PR.

jgabry:
Actually I had some time earlier than I though. This looks good. I'm going to add an item to NEWS.md and then merge. Thanks @Ozan147! 

I think eventually we should consider adding something similar for the `ppc_` functions (e.g., yrep needs to be a matrix but we could convert to matrix from draws_rvar and other formats) but we should wait to work on that until after #222 is merged.

jgabry:
I'm also going to add something to the doc about supporting posterior formats and then I'll merge this

PR #278, opened 2021-10-07
Add mcse plots similar to neff and rhat plots

jgabry:
This PR is not new, just a reworked version of PR #210 from @VeenDuco. For some reason that PR got messed up and it lists 223 files changed when there should only be a few. This separate PR includes what I think were the intended changes from @VeenDuco, which I had already looked at and approved (so I think this is ready to go). I also added a few tests. 

@VeenDuco would you mind taking a quick look at this before I merge it?

### DIFF
--- a/tests/testthat/test-helpers-mcmc.R
+++ b/tests/testthat/test-helpers-mcmc.R
@@ -299,20 +299,20 @@ test_that("tidy parameter selection throws correct errors", {
 # rhat and neff helpers ---------------------------------------------------
 test_that("diagnostic_factor.rhat works", {
   rhats <- new_rhat(c(low = 0.99, low = 1, low = 1.01,
-                      ok = 1.06, ok = 1.09, ok = 1.1,
+                      mid = 1.06, mid = 1.09, mid = 1.1,
                       high = 1.2, high = 1.7))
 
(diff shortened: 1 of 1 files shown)


VeenDuco:
Hey @jgabry, I think most looks good. Only in `R/mcmc-diagnostics.R` there are still some occurrences of `  labels = c("low", "ok", "high")` instead of `  labels = c("low", "mid", "high")`. I think everywhere else `ok` was replaced with `mid`. Everything else is a quite straightforward adaptation of the other functions and looks good.

ishaan-arora-1:
Hey @jgabry, what is the reason that this pr was not merged? i was trying to look into this direction for contributing.

jgabry:
I honestly don't remember!

ishaan-arora-1:
@jgabry how would you like this being a part of the summer project?

jgabry:
> how would you like this being a part of the summer project?

Sure, although I think not a big part for two reasons: 

- There have been 0 requests for this plot in the years since this PR was opened but not merged
- I think the work on this should be mostly finished already, right? This branch would just need to updated to fix merge conflicts and make any other updates that affect diagnostic plots.

PR #292, opened 2022-11-12
Handle deprecation of `aes_` (ggplot2) and `.data` (tidyselect)

jgabry:
`aes_` is now deprecated in ggplot2

`.data` is deprecated for tidyselect but still used for data masking


jgabry:
Passing except for a vignette failure on windows that isn't related to this PR (there's still an issue running rstan via GitHub actions on windows)

PR #299, opened 2023-03-06
Check intervals numeric

tony-stone:
Setting `prob <- FALSE`, e.g. in `mcmc_intervals`, does not raise an error but has unintended consequences.


jgabry (approved):
Thank you!

PR #300, opened 2023-03-16
add 'bins' argument to many histogram plots

paul-buerkner:
This PR adds the `bins` argument to a lot of histogram plots. So far, you could only customize the histogram bins via `binwidth` and `breaks`, but the often most simple and convenient option to just specify the number of bins via the `bins` argument of `ggplot2::geom_histogram` was missing so far. This PR resolves this issue.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,8 +1,8 @@
 Package: bayesplot
 Type: Package
 Title: Plotting for Bayesian Models
-Version: 1.10.0.9000
-Date: 2022-11-16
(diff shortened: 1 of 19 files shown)


jgabry:
I think the problem is that the most recent ggplot2 release fixed a bug in geom_hex that affects a few of our visual tests. I just updated the SVG files for the tests and hopefully the checks will pass now.

jgabry (approved):
Will merge if the checks pass now

paul-buerkner:
Thank you!

PR #307, opened 2023-10-25
better loo_pit_qq plots

### LINKED ISSUE #296: ppc_loo_pit_qq theoretical line
ppc_loo_pit_qq() currently adds a theoretical line that is not going from [0,0] to [1,1] but something else (maybe a line that passes through first and third quartiles?). When plotting pit values, the axes in the plot don't have a box corner at [0,0] and [1,1] and it is difficult to see where the line is going and it may look like as it would be the line from [0,0] to [1,1]. As in other rank ECDF plots the line is from [0,0] to [1,1], it would be if ppc_loo_pit_qq() would also plot a line from [0,0] to [1,1]

avehtari:
EDIT: fixes #296 

loo_pit_qq plots were using geom_qq + geom_qq_line, but that line was not useful in these plots. geom_qq_line is useful if the mean and scale of the data can be different and only the shape matters. When plotting PITs, they are constrained in range [0,1] and the qq's should follow the diagonal. When plotting with `compare='normal'`, the reference line should stiill be diagonal. This PR fixes the line and sets axis limit for the uniform qq-plot to be (0,1). Examples of old and new plots

Old uniform, as the line is diagonal it makes it more difficult to see the relevant difference:
image

New uniform:
image

Old normal:
image

New normal:
image

### DIFF
--- a/R/ppc-loo.R
+++ b/R/ppc-loo.R
@@ -189,7 +189,7 @@ ppc_loo_pit_overlay <- function(y,
       geom_line(
         aes(color = "y"),
         data = function(x) dplyr::filter(x, .data$is_y),
-        linewidth = 1,
+       linewidth = 1,
         lineend = "round",
         na.rm = TRUE) +
       scale_x_continuous(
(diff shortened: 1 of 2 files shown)


jgabry:
I think this is only failing because the visual tests detected the change to the plot. I will update the stored svg file and merge.

jgabry (approved):

PR #311, opened 2024-01-17
Prepare v1.11.0 

jgabry:
Branch for bayesplot v1.11.0

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,8 +1,8 @@
 Package: bayesplot
 Type: Package
 Title: Plotting for Bayesian Models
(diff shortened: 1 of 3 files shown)


jgabry:
There appears to be some change in R-devel that very slightly affects the svgs used for the visual tests. I don't think there's a way to get this to pass on release and devel because then we would need different sets of svgs. I think we'll just have to update the svgs after the current devel is released. 

We'll also need to update the visual tests after the next ggplot2 release since they are also affected by that.

jgabry:
Everything is passing except on r-devel for the reason explained above (the SVGs). Will merge and submit the release.

PR #285, opened 2022-06-13
Add option 'bins' to histogram functions

heavywatal:

PR #313, opened 2024-01-18
Add `bins` and `breaks` arguments to more histogram and hex plots

heavywatal:
This is a follow-up of #285 and #300.


jgabry:
Thank you! This looks good. The R CMD check failures are due to an unrelated issue that I'm working on fixing. Merging now.

jgabry (approved):

PR #329, opened 2024-07-31
Document problems with `ppc_stat` with `stat="mean"`

### LINKED ISSUE #321: Add warning about using ppc_stat stat="mean"
Most of the models do include parameters that do make the posterior predictive mean to match the observed mean, and the test statistic "mean" is perfectly dependent and not ancillary as required so that ppc would make some sense. `stat="mean"` might make sense given grouping if the groups don't have their own intercept parameters. I have recently seen too many times people posting `ppc_stat` plots with `stat="mean"` and the plots have zero information content. I guess we can't change the default to not break existing code.

- When calling `ppc_stat()` with the default `stat` or with explicit `stat="mean"`, create a warning saying "Test statistic "mean" is unlikely to be ancillary and thus likely to not provide any useful information. We recommend using some other test statistic." or if we want to avoid the word "ancillary" then warn: "Test statistic "mean" is very weak to detect anything interesting in most cases. We recommend using some other test statistic."
- Change the `ppc_stat` doc to describe the problem with "mean" test statistic and use some other test statistic in the example

jgabry:
closes #321

Adds section at the very top of the `ppc_stat` documentation (with **NOTE** in bold) that mentions the problem with `stat="mean"`: 

>   **NOTE:** Although the default test statistic is the mean, this is unlikely to detect anything interesting in most cases. In general we recommend using some other test statistic as discussed in Section 5 of Gabry et al. (2019).

### DIFF
--- a/R/ppc-test-statistics.R
+++ b/R/ppc-test-statistics.R
@@ -59,7 +59,7 @@
 #' @examples
 #' y <- example_y_data()
 #' yrep <- example_yrep_draws()
-#' ppc_stat(y, yrep)
+#' ppc_stat(y, yrep, stat = "median")
 #' ppc_stat(y, yrep, stat = "sd") + legend_none()
 #'
 #' # use your own function for the 'stat' argument
@@ -74,8 +74,8 @@
 #' # plots by group
 #' color_scheme_set("teal")
 #' group <- example_group_data()
-#' ppc_stat_grouped(y, yrep, group)
-#' ppc_stat_grouped(y, yrep, group) + yaxis_text()
+#' ppc_stat_grouped(y, yrep, group, stat = "median")
+#' ppc_stat_grouped(y, yrep, group, stat = "mad") + yaxis_text()
 #'
 #' # force y-axes to have same scales, allow x axis to vary
 #' ppc_stat_grouped(y, yrep, group, facet_args = list(scales = "free_x")) + yaxis_text()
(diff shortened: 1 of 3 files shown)


jgabry:
@avehtari If you prefer I can add a warning or message in addition to making this prominent in the documentation, although maybe making it very clear in the doc is enough?

avehtari:
I think many users who have already been using ppc_stat probably don't re-read the doc, so for those a warning would be good, but if you think that is too much, at least we should include the comment in the NEWS and release announcement, too

jgabry:
> I think many users who have already been using ppc_stat probably don't re-read the doc 

Yeah that's probably true. I've updated the PR to add a message if stat="mean".

jgabry:
> > I think many users who have already been using ppc_stat probably don't re-read the doc
> 
> Yeah that's probably true. I've updated the PR to add a message if stat="mean".

When running `ppc_stat`, `ppc_stat_2d`, `ppc_stat_freqpoly`, or the `_grouped` versions of those functions using `stat="mean"` the user will see the following: 

Screenshot 2024-08-02 at 12 26 45 PM

@avehtari Does that message look good to you? I think it's implied that they should therefore choose a different test statistic, so I don't think we need to say "we recommend choosing a different statistic". But do you want me to add something about where to find an explanation or this is enough?

avehtari:
I think this is fine. The documentation now refers to a text which explains things

PR #332, opened 2024-12-13
Ensure rank overlay plot starts at 0 even if not all bins present

### LINKED ISSUE #331: mcmc_rank_overlay: d_bin_counts doesn't contain rows for cases with n = 0
See also: https://discourse.mc-stan.org/t/rank-plots-dont-save-you-from-symmetric-non-stationary-chains/37624

In my case the first 4 bins all have count of 100 for the second chain, so the first chain's rank plot begins at 501. I suppose this could be a design choice but I think there should be bin entries in case n= 0.

image

jgabry:
fixes #331

Code to test: 

```r
x <- posterior::as_draws_array(
  list(
    list(theta = -2 + 0.003 * 1:1000 + arima.sim(list(ar = 0.7), n = 1000, sd = 0.5)),
    list(theta = 1 + -0.01 * 1:1000 + arima.sim(list(ar = 0.7), n = 1000, sd = 0.5))
  )
)
bayesplot::mcmc_rank_overlay(x)
```

### DIFF
--- a/R/mcmc-traces.R
+++ b/R/mcmc-traces.R
@@ -304,17 +304,24 @@ mcmc_rank_overlay <- function(x,
     mutate(cut = cut(.data$value_rank, n_bins)) %>%
     group_by(.data$cut) %>%
     mutate(bin_start = min(.data$value_rank)) %>%
-    ungroup() %>%
-    select(-c("cut"))
+    ungroup()
 
+  # Count how many values fall into each bin per chain & parameter
   d_bin_counts <- data %>%
     left_join(histobins, by = "value_rank") %>%
     count(.data$parameter, .data$chain, .data$bin_start)
 
+  # Now ensure that all combinations of parameter, chain, and bin_start exist
+  # even if no counts are present (https://github.com/stan-dev/bayesplot/issues/331)
+  all_params_chains <- dplyr::distinct(data, .data$parameter, .data$chain)
+  all_bins <- dplyr::distinct(histobins, .data$bin_start, .data$cut)
+  combos <- dplyr::cross_join(all_params_chains, all_bins)
+  d_bin_counts <- full_join(combos, d_bin_counts, by = c("parameter", "chain", "bin_start")) %>%
+    mutate(n = dplyr::coalesce(n, 0L))
+
   # Duplicate the final bin, setting the left edge to the greatest x value, so
   # that the entire x-axis is used,
   right_edge <- max(data$value_rank)
-
   d_bin_counts <- d_bin_counts %>%
     dplyr::filter(.data$bin_start == max(.data$bin_start)) %>%
     mutate(bin_start = right_edge) %>%
--- a/R/mcmc-traces.R
+++ b/R/mcmc-traces.R
@@ -313,11 +313,15 @@ mcmc_rank_overlay <- function(x,
 
   # Now ensure that all combinations of parameter, chain, and bin_start exist
   # even if no counts are present (https://github.com/stan-dev/bayesplot/issues/331)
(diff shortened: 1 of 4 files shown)


R/mcmc-traces.R
```diff
@@ -304,17 +304,24 @@ mcmc_rank_overlay <- function(x,
...
     count(.data$parameter, .data$chain, .data$bin_start)
 
+  # Now ensure that all combinations of parameter, chain, and bin_start exist
+  # even if no counts are present (https://github.com/stan-dev/bayesplot/issues/331)
+  all_params_chains <- dplyr::distinct(data, .data$parameter, .data$chain)
+  all_bins <- dplyr::distinct(histobins, .data$bin_start, .data$cut)
+  combos <- dplyr::cross_join(all_params_chains, all_bins)
+  d_bin_counts <- full_join(combos, d_bin_counts, by = c("parameter", "chain", "bin_start")) %>%
+    mutate(n = dplyr::coalesce(n, 0L))
+
```
jgabry: There may be a more elegant solution, but this seems to get the job done

jgabry:
The failure on macOS-latest (devel) is due to the vdiffr package failing to compile on GitHub actions

sims1253:
Maybe add a `set.seed` to the test to make it deterministic?

R/mcmc-traces.R
sims1253: Not sure if this is more elegant but an expand.grid with one left join and an if_else to fill the zeros should also work. Something like the following. Reads a little easier for me, though that might just be me. (not tested!)
  ``` 
  all_combos <- tibble::as_tibble(expand.grid(
    parameter = unique(data$parameter),
    chain = unique(data$chain),
    bin_start = unique(histobins$bin_start),
    stringsAsFactors = FALSE
  ))

  d_bin_counts <- all_combos %>%
    left_join(d_bin_counts, by = c("parameter", "chain", "bin_start")) %>%
    mutate(n = if_else(is.na(n), 0L, n))
  ```

sims1253:
```
# Ensure all combinations exist and count values
d_bin_counts <- data %>%
  left_join(histobins, by = "value_rank") %>%
  count(.data$parameter, .data$chain, .data$bin_start) %>%
  right_join(
    dplyr::expand(.data$parameter, .data$chain, unique(histobins$bin_start)),
    by = c("parameter", "chain", "bin_start")
  ) %>%
  mutate(n = dplyr::coalesce(n, 0L))
  ```
  Or courtesy of claude, telling me my solution was verbouse and not dplyr-ish enough.

jgabry:
> Maybe add a `set.seed` to the test to make it deterministic?

There's a `set.seed` in the file where I added the code to create the data for the test (it just doesn't show up in the diff since I didn't change it): 

https://github.com/stan-dev/bayesplot/blob/414755546305ea2ef06f70145c78712a024e8963/tests/testthat/data-for-mcmc-tests.R#L33

R/mcmc-traces.R
jgabry: Thanks to you and Claude, both!

jgabry: > dplyr::expand

Except Claude didn't realize there is no `dplyr::expand` . There is `tidyr::expand`, but we don't already depend on `tidyr` so probably best not to add another dependency just for this.

jgabry:
@sims1253 Thanks. I just pushed a commit that uses basically your first suggestion of `expand_grid` plus a `left_join`.

R/mcmc-traces.R
sims1253: Ha, great help :D

PR #338, opened 2025-01-30
Update GitHub actions

jgabry:
Update to latest GHA workflows


jgabry:
I'm not sure why some of the visual tests are failing on `oldrel-1`, but the differences in the SVGs are imperceptible to the eye, so I'm not concerned.

PR #341, opened 2025-04-08
Prepare v1.12.0 for release

jgabry:
We need to release bayesplot before the next ggplot2 to avoid some test failures on CRAN.

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -4,9 +4,11 @@ on:
   push:
     branches: [main, master]
   pull_request:
-    branches: [main, master, patch-for-cran]
+    branches: [main, master]
 
-name: R-CMD-check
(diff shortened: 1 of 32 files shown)


TeemuSailynoja:
What is the timeline on this? I see the ggplot2 release Issue and PR were created yesterday, but they seem to be just waiting for approval from CRAN, so I imagine this bayesplot release is also on a very expedited schedule?

jgabry:
Yeah I think we should do it this week. If you have things to add to bayesplot that aren’t ready yet I’m happy to do another release very soon after this one.

jgabry:
- The failure on Ubuntu with the old version of R is due to small differences in the SVGs for the visual tests. I can't tell the difference by eye. 

- Currently running reverse dependency checks. If no issues I will go ahead and merge this PR and submit the release.

PR #337, opened 2025-01-30
Compatibility with ggplot2 3.6.0

teunbrand:
Hi there,

Apologies for not posting an issue first.
The ggplot2 package is planning an update for around May 2025 and a reverse dependency test identified a problem with the bayesplot package.
The details are explained in https://github.com/tidyverse/ggplot2/issues/6290, but essentially ggplot2 doesn't populate the `plot$labels` field before plot building anymore, which violates some test assumptions in this package.

This PR changes the tests to be compatible with both versions of ggplot2. 
In addition, some deprecated functionality is avoided.
You can test the changes yourself with the development version of ggplot2 (`pak::pak("tidyverse/ggplot2")`).
There are some visual changes as well, but I'll leave that up to you to decide wether intervention is needed.

Best,
Teun

PR #345, opened 2025-04-23
Add ppc_loo_pit_ecdf

TeemuSailynoja:
Made an implementation of the `ppc_pit_ecdf()` for LOO-PITs. This is mostly a convenience function, as one could already pass LOO-PITs to the `ppc_pit_ecdf` in its `pit` argument. Now works directly with for example `brms::pp_check(brms_fit, "loo_pit_ecdf")`

I have two questions:
1.  ppc_loo_pit_ecdf() uses a `prob` argument. `ppc_loo_pit_intervals` already imports the `args-prob-prob_outer` template, which has a different doc string for `prob`. This template is used outside the ppc_loo functions, so I don´t think adding text specific to this new function there is the best solution. How should this be handled?
2. I made `ppc_loo_pit_ecdf` have a slightly different signature to `ppc_pit_ecdf`. Namely, instead of `K`, this new function uses a more informative `eval_points`, and instead of the boolean `interpolate_adj`, now I have `interval_method = c("interpolate", "optimize")`. Should I hold on to this change and have the functions share their signature? I think the new one is more clear and `ppc_pit_ecdf` should also move towards that eventually.


jgabry:
> 1. ppc_loo_pit_ecdf() uses a `prob` argument. `ppc_loo_pit_intervals` already imports the `args-prob-prob_outer` template, which has a different doc string for `prob`. This template is used outside the ppc_loo functions, so I don´t think adding text specific to this new function there is the best solution. How should this be handled?

Maybe we don't use the `args-prob-prob_outer` template for `ppc_loo_pit_intervals` and instead you can do something like this?  

    @param prob For `ppc_loo_pit_intervals()` `prob` is used for X. For `ppc_loo_pit_ecdf()` `prob` is used for Y.`

> 2. I made ppc_loo_pit_ecdf have a slightly different signature to ppc_pit_ecdf. Namely, instead of K, this new function uses a more informative eval_points, and instead of the boolean interpolate_adj, now I have interval_method = c("interpolate", "optimize"). Should I hold on to this change and have the functions share their signature? I think the new one is more clear and ppc_pit_ecdf should also move towards that eventually.

It would be nice to have the same signature, but I agree with you that your new one is more clear. Since the new one is better, it's ok with me to have a different signature for now until we can also update the one for `ppc_pit_ecdf`.

man/PPC-loo.Rd
```diff
@@ -175,26 +175,30 @@ compares computed PIT values to the standard uniform distribution. If
...
+faster. If \code{pit} is supplied, defaults to \code{length(pit)}, otherwise
+\code{yrep} determines the maximum accuracy of the estimated PIT values and
+\code{K} is set to \code{min(nrow(yrep) + 1, 1000)}.}
 
-\item{prob}{For \code{ppc_loo_pit_ecdf()}, the desired simultaneous coverage
-level of the bands around the ECDF. A value in (0,1).}
+\item{prob, prob_outer}{Values between \code{0} and \code{1} indicating the desired
+probability mass to include in the inner and outer intervals. The defaults
+are \code{prob=0.5} and \code{prob_outer=0.9} for \code{ppc_loo_intervals()} and
+\code{prob = 0.99} for \code{ppc_loo_pit_ecdf()}.}
```
TeemuSailynoja, line 188: I removed the use of the documentation template and just list the default of `ppc_loo_pit_ecdf()` separately.

man/PPC-loo.Rd
```diff
@@ -62,10 +62,10 @@ ppc_loo_pit_ecdf(
   ...,
   psis_object = NULL,
   pit = NULL,
-  eval_points = NULL,
+  K = NULL,
   prob = 0.99,
   plot_diff = FALSE,
-  interval_method = c("interpolate", "optimize")
+  interpolate_adj = NULL
```
TeemuSailynoja: I also unified the arguments for now, to keep the code more simple.
I suggest that, when I work on the documentation and possibly need to rearrange how the PPC function pages are grouped, I will add the clearer argument names and add a deprecation warning to the old ones.

TeemuSailynoja: Also, with this rearranging, the template could come back to use with `ppc_loo_pit_intervals()`.

It would probably make sense to have `ppc_(loo)_pit_ecdf()`, `ppc_loo_pit_qq()`, and `ppc_loo_pit_overlay()` share a documentation page, and move `ppc_loo_intervals()`, and `ppc_loo_ribbon()` join their LOOless counterparts on the PPC intervals documentation page.

man/PPC-loo.Rd
jgabry: Looks good

man/PPC-loo.Rd
jgabry: > I also unified the arguments for now, to keep the code more simple.
I suggest that, when I work on the documentation and possibly need to rearrange how the PPC function pages are grouped, I will add the clearer argument names and add a deprecation warning to the old ones.

Sounds good, I like that plan.

jgabry: > Also, with this rearranging, the template could come back to use with `ppc_loo_pit_intervals()`.
> 
> It would probably make sense to have `ppc_(loo)_pit_ecdf()`, `ppc_loo_pit_qq()`, and `ppc_loo_pit_overlay()` share a documentation page, and move `ppc_loo_intervals()`, and `ppc_loo_ribbon()` join their LOOless counterparts on the PPC intervals documentation page.

That also sounds good. When I first included those LOO-based plots I thought it made sense to put them all together. But I think you're right that this reorganization makes sense at this point. I think we'll just want to link between the documentation pages, pointing out that there are other LOO-based plots in various locations.

tests/testthat/test-ppc-loo.R
```diff
@@ -73,10 +73,24 @@ test_that("ppc_loo_pit_qq returns ggplot object", {
...
+  skip_if_not_installed("rstanarm")
+  skip_if_not_installed("loo")
+  expect_gg(p0 <- ppc_loo_pit_ecdf(y, yrep, lw))
+  expect_gg(p1 <- ppc_loo_pit_ecdf(y, yrep, lw))
+  expect_gg(p2 <- ppc_loo_pit_ecdf(y, yrep, psis_object = psis1))
+  expect_equal(p1$labels$x, "LOO PIT")
+  expect_equal(p1$labels$y, "ECDF")
+  expect_equal(p1$data, p2$data)
+  expect_gg(p3 <- ppc_loo_pit_ecdf(y, yrep, lw, plot_diff = TRUE))
+  expect_equal(p3$labels$y, "ECDF difference")
```
jgabry: I think for future compatibility with ggplot2 we probably need to do what was done in https://github.com/stan-dev/bayesplot/pull/337/files#diff-d31e64955d39a8924be5d979cb21c5e96514987aeff9ca8cf2af067e98f3b156 with checking for the `get_labs()` function  and using it if found.

jgabry: (I just noticed that I haven't done this for the tests for `ppc_loo_pit_qq` either. If you want you can do it with this PR but I can also do it separately since it's not related to your PR)

jgabry (commented):
The function works great. I made one small comment on the tests and then there's mentioning the function in the "Plot Descriptions" section of the doc, but otherwise I think this is ready without any other changes. 

One strange thing is that two of the visual tests are failing for me locally but apparently not on Github Actions. But they're not your tests for the new function, they're `ppc-loo-pit-overlay-default.svg` and `ppc-loo-pit-qq-default.svg`. Does everything pass for you locally?

tests/testthat/test-ppc-loo.R
TeemuSailynoja: I can also add the part for `ppc_loo_pit_qq`.

jgabry (approved):
Looks good, thanks. I figured out why those tests were failing for me. I had a few development versions of various dependencies installed that subtly affected the SVGs for some reason. Not a big deal. Everything passes for me now. 

Merging now.

PR #347, opened 2025-04-29
Add possibility for left-truncation to ppc_km_overlay() and ppc_km_overlay_grouped()

### LINKED ISSUE #290: ppc_km_overlay(): add argument for left-truncation?
*Summary*
It would be great if `ppc_km_overlay()` could accomodate left-truncation when computing the Kaplan-Meier functions for the original sample and for the simulated data from the PPD, and not only right-censoring.

Would you consider adding a `truncation_y` (?) argument to `ppc_km_overlay()` to pass the left-truncation variable in the original sample to `survival::Surv()`?

*Example*
Weibull survival times, observed conditionally on truncation times. 

Standard plot produced by `ppc_km_overlay()` together with KM estimates that take into account left-truncation (red, solid line) and true survival curve (black, dashed line) for the original sample (note that the `ppc_km_overlay()`'s KM curves for the simulated PPD data ignore truncation too, but the underlying simulated times via `brms::posterior_predict()` are conditional on the observed truncation times).

<a href="https://freeimage.host/i/p8pfhG"></a>

```
library(brms)
library(survival)
library(ggplot2)

simdata <- local({
  set.seed(1901)
  N <- 1000
  y <- rweibull(N, shape = 1.2, scale = 1) 
  v <- rexp(N)
  data.frame(time = y, v = v)
})
simdata_trunc <- simdata[simdata$time >= simdata$v, ]

formula_brms <- bf(time | trunc(lb = v, ub = Inf) ~ 1,
                   family = weibull())
fit_brms <- brm(formula_brms,
                data = simdata_trunc,
                chains = 1,
                iter = 2000,
                seed = 12345,
                backend = "cmdstanr",
                refresh = 0)

set.seed(9)
pp <- posterior_predict(fit_brms, draw_ids = 1:5)
ppc <- bayesplot::ppc_km_overlay(simdata_trunc$time, pp, status_y = rep(1, nrow(simdata_trunc)))
km_trunc <- survfit(Surv(v, time, rep(1, nrow(simdata_trunc))) ~ 1, data = simdata_trunc)

ppc +
  geom_line(data = data.frame(time = km_trunc$time, surv = km_trunc$surv),
            aes(x = time, y = surv), col = "red",
            inherit.aes = FALSE, ,lwd = 1) +
  geom_function(fun = pweibull, args = list(lower.tail = FALSE, shape = 1.2, scale = 1),
                col = "black", lty = 2, inherit.aes = FALSE)
```

Sakuski:
Fixes #290


R/ppc-censoring.R
```diff
@@ -70,6 +70,7 @@ ppc_km_overlay <- function(
   yrep,
   ...,
   status_y,
+  left_truncation_y = NULL,
```
jgabry, line 89: Can you add documentation for this argument? You can put it after the documentation for the `status_y` argument. You just need to do something like 

```
#' @param left_truncation_y Description of the purpose of the argument, the default setting, and the required length. 
```

It would also be great to add something simple to the Examples section in the documentation.

Sakuski: I had wrongly assumed in my implementation that time > 0. This is now fixed and the tests pass. I also added documentation and an example for the new parameter. I am away for the rest of the week but next week I can create tests for the new functionality. @TeemuSailynoja promised to help me with those if needed. Is there anything else you would like me to do?

jgabry:
Thanks! Sorry for the delay in getting back to you, I've been traveling. I'll be able to take a look again hopefully tomorrow (or possibly later this week).

R/ppc-censoring.R
```diff
@@ -58,18 +58,29 @@
 #' \donttest{
 #' ppc_km_overlay_grouped(y, yrep[1:25, ], group = group, status_y = status_y)
 #' }
+#' # With left-truncation (delayed entry) times:
+#' left_truncation_y <- runif(length(y), min = 0, max = 0.6) * y
+#' \donttest{
+#' ppc_km_overlay(y, yrep[1:25, ], status_y = status_y,
+#'               left_truncation_y = left_truncation_y)
+#' }
```
jgabry: I get a warning when I try this example: 

```
Warning message:
In survival::Surv(time = left_truncation_y[data$y_id], time2 = data$value,  :
  Stop time must be > start time, NA created
```

jgabry (commented):
Sorry again for the delay in looking at this again. This looks great. Thanks for adding the tests. I pushed a couple of minor commits related to the tests (the random number generation was affecting some of the data for other tests unrelated to this, which wasn't your fault).  

I think this is basically ready to be merged now, except there's just one thing I noticed that may or may not be an issue. I get a warning when I run the example you added to the documentation: 

```
> # With left-truncation (delayed entry) times:
> left_truncation_y <- runif(length(y), min = 0, max = 0.6) * y
> 
> ppc_km_overlay(y, yrep[1:25, ], status_y = status_y,
+                left_truncation_y = left_truncation_y)
Warning message:
In survival::Surv(time = left_truncation_y[data$y_id], time2 = data$value,  :
  Stop time must be > start time, NA created
```

Is that to be expected?

R/ppc-censoring.R
Sakuski: The warning was to be expected in case of that example, but I made a cleaner example that does not produce warnings. In addition, I think the new example visualizes the effect of left-truncation a bit better.

Sakuski: In real life, the warning is an indication of an impossible observation or posterior predictive draw, so the warning itself is good. When creating examples, we just have to make sure that each value in left_truncation_y is lower than the corresponding value in y or corresponding values in yrep so we don't get the warning.

jgabry: Ok great, thanks, that makes sense.

jgabry (approved):
I think this is ready to merge. I’ll let the tests finish running and if everything passes I’ll go ahead and merge it later tonight or tomorrow. Thanks again for the PR

Sakuski:
@jgabry Before you merge, I think you should generate a new .Rd file for PPC-censoring, since I changed the example in my last commit. I didn't know how to do that.

jgabry:
> @jgabry Before you merge, I think you should generate a new .Rd file for PPC-censoring, since I changed the example in my last commit. I didn't think of that yesterday.

Will do. And actually I just noticed one more issue. The object `min_vals` in the example code is never defined anywhere so it errors.

R/ppc-censoring.R
```diff
@@ -59,7 +59,11 @@
 #' ppc_km_overlay_grouped(y, yrep[1:25, ], group = group, status_y = status_y)
 #' }
 #' # With left-truncation (delayed entry) times:
-#' left_truncation_y <- runif(length(y), min = 0, max = 0.6) * y
+#' condition <- y > mean(y) / 2
+#' left_truncation_y[condition] <- pmin(
+#'   runif(sum(condition), min = 0.6, max = 0.99) * y[condition],
+#'   min_vals[condition] - 0.001
```
jgabry, line 67: This `min_vals` object isn't defined anywhere

jgabry (changes requested):
The last thing before we merge is to define the `min_vals` object in the example

R/ppc-censoring.R
Sakuski: It turns out I had not included two lines of my example code there. This is now fixed.

jgabry (approved):
Thanks, looks good now. I just regenerated the Rd file and triggered the unit tests to run again. Assuming everything passes (I assume it will) we can merge this!

jgabry:
Merging now, thanks again!

Issue #348, opened 2025-05-13
`ppc_error_scatter_avg_vs_x`, has unstable residuals when the noise distribution has heavy tails; needs median not mean?


kruschke:
`brms::pp_check(type = "error_scatter_avg_vs_x")`, which calls `bayesplot::ppc_error_scatter_avg_vs_x`, has residuals that vary wildly from one run to the next when the noise distribution (`family`) is a Student $t$ distribution with small `nu` (aka `df`).

I suspect the problem is caused by the "average" being computed as the *mean*, which gets wildly distorted by outliers generated by the kurtotic noise distribution. I suspect the problem would be greatly ameliorated if the average could instead be computed as the *median*. But there seems to be no option for this, while some of the `brms` functions do have an argument `robust = TRUE`.

The linked HTML (and .Rmd) file has two examples, identical to each other except the first example uses data from a highly kurtotic $t$ distribution, while the second example used data from an essentially normal $t$ distribution.
HTML: https://drive.google.com/file/d/1aGwO9i7RuXQoVVzkHkV4wC02uqKs43ys/view?usp=drive_link
.Rmd: https://drive.google.com/file/d/1ksRr_2VOgMornK3s_Dk5sA0xoKPOywRA/view?usp=drive_link

tjmahr:
Thanks for the issue. (Fan of your book with the dogs!) 

I submitted a patch #349 to let users override the mean function via `fun_avg` argument.

jgabry:
Thanks @kruschke, I agree we should provide an option to for using other functions for averaging. Thanks @tjmahr for the PR. Will review soon.

PR #349, opened 2025-05-13
add fun_avg to ppc_avg functions

tjmahr:
Following #348, allow user to set the averaging function so that, e.g., user can choose median for heavy-tailed distributions.

``` r
l <- jsonlite::read_json(
  "https://gist.githubusercontent.com/tjmahr/3eba4f51e4b122b19417718b56423a6c/raw/31711cacef59eff69ed95eb51e2ae844eaad01b7/draws.json", 
  simplifyVector = TRUE
)

# Defaults to mean
bayesplot::ppc_error_scatter_avg_vs_x(l$y, l$yrep, l$x) +
  ggplot2::expand_limits(x = c(-10, 10))
```

``` r
bayesplot::ppc_error_scatter_avg_vs_x(l$y, l$yrep, l$x, fun_avg = median) + 
  ggplot2::expand_limits(x = c(-10, 10))
```

<sup>Created on 2025-05-13 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>


R/ppc-errors.R
```diff
@@ -10,6 +10,8 @@
 #' @template args-group
 #' @template args-facet_args
 #' @param ... Currently unused.
+#' @param fun_avg Function to apply to compute the posterior average.
+#'   Defaults to `"mean"`.
```
TeemuSailynoja: In the `ppc_stat` functions, we have a similar argument, by the name `stat`, which does a very similar job:
> #' @param stat A single function or a string naming a function, except for the
#'   2D plot which requires a vector of exactly two names or functions. In all
#'   cases the function(s) should take a vector input and return a scalar
#'   statistic. If specified as a string (or strings) then the legend will
#'   display the function name(s). If specified as a function (or functions)
#'   then generic naming is used in the legend.

We could align this doc to read, for example:
> #' @param fun_avg A function or a string naming a function for computing the
#'   posterior average. In both cases, the function should take a vector input and
#'   return a scalar statistic. If specified as a string, then the legend will
#'   display the function name. If specified as a function
#'   then generic naming is used in the legend.
#'       Defaults to `"mean"`.

R/ppc-scatterplots.R
```diff
@@ -11,6 +11,8 @@
 #' @template args-group
 #' @template args-facet_args
 #' @param ... Currently unused.
+#' @param fun_avg Function to apply to compute the posterior average.
+#'   Defaults to `"mean"`.
```
TeemuSailynoja: same as above.

TeemuSailynoja (commented):
Thank you for the fast reaction. This looks like a clean implementation to address the listed issue. My only comment is about the documentation, which could mention the expected format of `fun_avg`.

avehtari:
I was also thinking `stat` instead of `fun_avg`. 

It would be good to change the y-axis label to show the `stat` which was used, and also add parentheses. For example: 
"median(y - y_rep)" or "y - median(y_rep)"

R/ppc-errors.R
tjmahr: The `Average y - y_rep` axis label is not affected. I didn't want to make `yrep_avg_label()` and `error_avg_label()` depend on `fun_avg` or change the default "Average y - y_rep" labels.

It does affect the `$rep_label` in `ppc_scatter_avg_data(y, yrep)` when fun_avg is a string.

tjmahr: wait, i'm just noticing Aki's comment. I'll switch to stat.

kruschke:
Following up on @avehtari , also would be good to change the x-axes label to show the `x` variable name instead of just generic $x$.

R/ppc-errors.R
jgabry: Thanks @tjmahr. The code looks good. And I agree with changing to `stat`

tjmahr:
1. I have changed `fun_avg` to have the name `stat`.
2. The stat name is shown on the axis label.
3. I have made the `stat` argument handle functions given by strings, function objects, anonymous function objects, and quoted functions names. 

Because the function name is used for both labeling and invoking a function name, I added `as_tagged_function()` that keeps track of the expression provided by the user.

*** 

``` r
library(bayesplot)
#> This is bayesplot version 1.12.0.9000

l <- jsonlite::read_json(
  "https://gist.githubusercontent.com/tjmahr/3eba4f51e4b122b19417718b56423a6c/raw/31711cacef59eff69ed95eb51e2ae844eaad01b7/draws.json", 
  simplifyVector = TRUE
)
y <- l$y 
yrep <- l$yrep 
x <- l$x
```

stat is inlined into the axis label

``` r
ppc_scatter_avg(y, yrep, stat = "mean")
```



stat can be a function

``` r
ppc_scatter_avg(y, yrep, stat = median)
```



or a symbol

``` r
ppc_scatter_avg(y, yrep, stat = quote(median))
```



stat can be a primitive function

``` r
ppc_error_scatter_avg_vs_x(y, yrep, x, stat = min)
```



stat can be an anonymous function, but it gets a generic label

``` r
ppc_scatter_avg(y, yrep, stat = function(x) quantile(x, .1))
```



<sup>Created on 2025-05-14 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

TeemuSailynoja:
> Following up on @avehtari , also would be good to change the x-axis label to show the `x` variable name instead of just generic "$x$".

If I'm not wrong, this is more challenging, as when passing `x = data_frame$col_name`, the information of the name of the column is lost, and only a numeric vector is passed to `x`.

tjmahr:
> If I'm not wrong, this is more challenging, as when passing `x = data_frame$col_name`, the information of the name of the column is lost, and only a numeric vector is passed to `x`.

No, you can capture that. (`plot()` does that.) It's really easy in a single function setup.

``` r
library(rlang)
library(ggplot2)
f <- function(x, y) {
  qx <- enquo(x)
  qy <- enquo(y)
  data <- data.frame(x = x, y = y)
  ggplot(data) + 
    aes(x, y) + 
    geom_point() + 
    labs(
      x = as_label(quo_get_expr(qx)),
      y = as_label(quo_get_expr(qy))
    )
}

data <- data.frame(
  y = bayesplot::example_y_data(),
  x = bayesplot::example_x_data()
)
f(data$x, data$y)
```



<sup>Created on 2025-05-15 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

The hard part for us is that when you call `f(x = my_x, y = my_y)` which internally calls `g(x, y)`, then g() loses access to the user expression unless you take extra care along the way. (Or use quosures. I'm not sure yet.) We do that when we modify data and call some internal `pp_()` function to render the plot.

My update yesterday kept track of the expression for `stat` along the way but it was pretty tricky.

tjmahr:
1. Label anonymous functions with "stat"
2. Support ~ style anonymous functions
3. Simplify as_tagged_function() by using quosures

*** 

``` r
library(bayesplot)
#> This is bayesplot version 1.12.0.9000

l <- jsonlite::read_json(
  "https://gist.githubusercontent.com/tjmahr/3eba4f51e4b122b19417718b56423a6c/raw/31711cacef59eff69ed95eb51e2ae844eaad01b7/draws.json", 
  simplifyVector = TRUE
)
y <- l$y 
yrep <- l$yrep 
x <- l$x
```

Function name

``` r
ppc_scatter_avg(y, yrep, stat = "mean")
```



``` r
ppc_error_scatter_avg(y, yrep, stat = "mean")
```



Function object

``` r
ppc_scatter_avg(y, yrep, stat = median)
```



``` r
ppc_error_scatter_avg(y, yrep, stat = median)
```



Primitive function

``` r
ppc_scatter_avg(y, yrep, x, stat = min)
```



``` r
ppc_error_scatter_avg_vs_x(y, yrep, x, stat = min)
```



Anonymous function

``` r
ppc_scatter_avg(y, yrep, x, stat = function(x) quantile(x, .1))
```



``` r
ppc_error_scatter_avg_vs_x(y, yrep, x, stat = function(x) quantile(x, .1))
```



Anonymous function (formulas)

``` r
ppc_scatter_avg(y, yrep, x, stat = ~ quantile(.x, .1))
```



``` r
ppc_error_scatter_avg_vs_x(y, yrep, x, stat = ~ quantile(.x, .1))
```



<sup>Created on 2025-05-15 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

tjmahr:
This patch should be ready for review.

The expression for `x` now appears in the axis label. 

``` r
library(bayesplot)
#> This is bayesplot version 1.12.0.9000
l <- jsonlite::read_json(
  "https://gist.githubusercontent.com/tjmahr/3eba4f51e4b122b19417718b56423a6c/raw/31711cacef59eff69ed95eb51e2ae844eaad01b7/draws.json", 
  simplifyVector = TRUE
)

ppc_error_scatter_avg_vs_x(l$y, l$yrep, l$x, stat = "sd")
```



<sup>Created on 2025-05-16 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

kruschke:
Sorry if I'm late to the party, but shouldn't the residual label be $y - stat(y_{rep})$, not $stat(y - y_{rep})$? Isn't $y - stat(y_{rep})$ being computed internally, not $stat(y - y_{rep})$?

tjmahr:
@kruschke: 

> But the related thread regarding residuals https://github.com/stan-dev/bayesplot/pull/349 seems to suggest that residuals are computed as s t a t ( y − y r e p ) , with s t a t ( y r e p ) not separately, explicitly computed. Hmmm...?

> Sorry if I'm late to the party, but shouldn't the residual label be $y - stat(y_{rep})$, not $stat(y - y_{rep})$? Isn't $y - stat(y_{rep})$ being computed internally, not $stat(y - y_{rep})$?

stat(y - y_rep) is what is computed internally.

DESCRIPTION
```diff
@@ -26,7 +26,7 @@ URL: https://mc-stan.org/bayesplot/
 BugReports: https://github.com/stan-dev/bayesplot/issues/
 SystemRequirements: pandoc (>= 1.12.3), pandoc-citeproc
 Depends:
-    R (>= 3.1.0)
+    R (>= 4.1.0)
```
jgabry, line 29: I think at this point it's been out long enough that bumping the required R version is fine.

R/bayesplot-helpers.R
```diff
@@ -469,3 +469,53 @@ grid_lines_y <- function(color = "gray50", size = 0.2) {
...
+# Resolve a function name and store the expression passed in by the user
+#' @noRd
+#' @param f a function-like thing: a string naming a function, a function
+#'   object, an anonymous function object, a formula-based lambda, and `NULL`.
+#' @param fallback character string providing a fallback function name
+#' @return the function named in `f` with an added `"tagged_expr"` attribute
+#' containing the expression to represent the function name and an
+#' `"is_anonymous_function"` attribute to flag if the expression is a call to
+#' `function()`.
+as_tagged_function <- function(f = NULL, fallback = "func") {
```
jgabry, line 484: This function works as described, just one question: how come sometimes you call rlang functions with `rlang::foo()` and other times just `foo()`? I think either way works since we have `@import rlang` in `bayesplot-package.R`, just curious if your choices were intentional.

R/ppc-scatterplots.R
```diff
@@ -31,10 +36,10 @@
 #'   }
 #'   \item{`ppc_scatter_avg()`}{
 #'    A single scatterplot of `y` against the average values of `yrep`, i.e.,
-#'    the points `(x,y) = (mean(yrep[, n]), y[n])`, where each `yrep[, n]` is
-#'    a vector of length equal to the number of posterior draws. Unlike
-#'    for `ppc_scatter()`, for `ppc_scatter_avg()` `yrep` should contain many
-#'    draws (rows).
+#'    the points `(x,y) = (average(yrep[, n]), y[n])`, where each `yrep[, n]` is
+#'    a vector of length equal to the number of posterior draws and `average()`
+#'    is summary statistic. Unlike for `ppc_scatter()`, for `ppc_scatter_avg()`
```
jgabry:
```suggestion
#'    is a summary statistic. Unlike for `ppc_scatter()`, for `ppc_scatter_avg()`
```

R/ppc-scatterplots.R
```diff
@@ -184,16 +196,19 @@ ppc_scatter_data <- function(y, yrep) {
...
   }
+  stat <- as_tagged_function({{ stat }})
 
-  data <- ppc_scatter_data(y = y, yrep = t(colMeans(yrep)))
+  data <- ppc_scatter_data(y = y, yrep = t(apply(yrep, 2, FUN = stat)))
   data$rep_id <- NA_integer_
-  levels(data$rep_label) <- "mean(italic(y)[rep]))"
+  levels(data$rep_label) <- yrep_avg_label(stat) |>
+    as.expression() |>
+    as.character()
```
jgabry, line 211: Fine by me to use the base R pipe, but we do import `%>%` from dplyr so could also use that.

jgabry (commented):
This looks good! I made a few minor comments/questions. Aside from those my only other question is does it makes sense to also use this new approach with the `stat` argument to `ppc_stat()`? You could do something similar to what you've done with the axis labels here but for the legend for `ppc_stat()`. What do you think?

R/bayesplot-helpers.R
tjmahr: Oops. Probably an artifact of me using typing `rlang::[TAB]` to use autocomplete to find the right rlang function.

jgabry:
@tjmahr Also see @avehtari's comment here https://github.com/stan-dev/bayesplot/issues/350#issuecomment-2902272870, which supports @kruschke's suggestion. When we were always using the mean it didn't matter if we were computing `y - stat(y_rep)` or `stat(y - y_rep)`. But if we're going to allow arbitrary functions then users will likely expect us to be computing `y - stat(y_rep)`, i.e. `y` minus a point prediction.

tjmahr:
> But if we're going to allow arbitrary functions then the two quantities won't necessarily be the same and users will likely expect us to be computing y - stat(y_rep), i.e. y minus a point prediction.

My thinking was stat(y - y_rep) is an average error, and y - stat(y_rep) is the error of the average prediction, so the former better matches the function names and how I think about ppc functions.

I don't oppose the change, and updating the axis label will make it unambiguous what's being computed.

jgabry:
Yeah I have mixed feelings about this. Both feel intuitive to me but represent slightly different things. 

A couple of options (there are probably others): 

1. Change `ppc_error_scatter_avg` to do `y - stat(y_rep)` on the y-axis and `stat(y_rep)` on the x-axis. This is what @kruschke was expecting and @avehtari agreed with, so it seems like many users would probably expect this. 

2. Keep `ppc_error_scatter_avg` as you have it and add a new one that does `y - stat(y_rep)` on the y-axis and `stat(y_rep)` on the x-axis. We could call it `ppc_residual()` or something like that. I think with the right documentation having both would probably be ok. 

Do you have a preference? Or a 3rd option?

tjmahr:
> Keep ppc_error_scatter_avg as you have it and add a new one that does y - stat(y_rep) on the y-axis and stat(y_rep) on the x-axis. We could call it ppc_residual() or something like that. I think with the right documentation having both would probably be ok.

Okay, that seems to be the plan now in #348.

tjmahr:
Here's the hopefully final set of test plots for the `stat` functionality.


``` r
library(bayesplot)
#> This is bayesplot version 1.13.0.9000

l <- jsonlite::read_json(
  "https://gist.githubusercontent.com/tjmahr/3eba4f51e4b122b19417718b56423a6c/raw/31711cacef59eff69ed95eb51e2ae844eaad01b7/draws.json", 
  simplifyVector = TRUE
)
y <- l$y 
yrep <- l$yrep 
x <- l$x
```

stat is inlined into the axis label

``` r
ppc_scatter_avg(y, yrep, stat = "mean")
```



stat can be a function

``` r
ppc_scatter_avg(y, yrep, stat = median)
```



stat can be a primitive function

``` r
ppc_error_scatter_avg_vs_x(y, yrep, x, stat = min)
```



stat can be an anonymous function, but it gets a generic label

``` r
ppc_scatter_avg(y, yrep, stat = function(x) quantile(x, .1))
```



``` r

ppc_scatter_avg(y, yrep, stat = ~quantile(.x, .1))
```



<sup>Created on 2025-07-02 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

tjmahr:
``` r
library(bayesplot)
#> This is bayesplot version 1.13.0.9000

l <- jsonlite::read_json(
  "https://gist.githubusercontent.com/tjmahr/3eba4f51e4b122b19417718b56423a6c/raw/31711cacef59eff69ed95eb51e2ae844eaad01b7/draws.json", 
  simplifyVector = TRUE
)
y <- l$y 
yrep <- l$yrep 
x <- l$x
```

stat is inlined into the axis label

``` r
ppc_scatter_avg(y, yrep, stat = "mean")
```



stat can be a function

``` r
ppc_scatter_avg(y, yrep, stat = median)
```



stat can be a primitive function

``` r
ppc_error_scatter_avg_vs_x(y, yrep, x, stat = min)
```



stat can be an anonymous function, but it gets a generic label

``` r
ppc_scatter_avg(y, yrep, stat = function(x) quantile(x, .1))
```



``` r

ppc_scatter_avg(y, yrep, stat = ~quantile(.x, .1))
```



<sup>Created on 2025-07-02 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

tjmahr:
This should be ready to merge.

Issue #343, opened 2025-04-14
Improve functions available for discrete observations


TeemuSailynoja:
Currently `bayesplot` offers are three functions for discrete PPCs: `ppc_bars`, `ppc_bars_grouped`, and `ppc_rootogram`. 

In [our preprint](https://teemusailynoja.github.io/visual-predictive-checks/#sec-visual-predictive-checks-for-binary-data), we argue why bar graphs are usually not a good PPC, especially for binary observations.

![Image](https://github.com/user-attachments/assets/d80d5d7c-7b0f-4a72-979e-5c0e0e196330)

## Calibration plots
I suggest adding the binned calibration plot and PAV-adjusted calibration plot under a function called `ppc_calibration`. For ordinal or categorical data, it would possibly be clearer to have another function that produces calibration plots for [1-vs-others](https://www.journalovi.org/2025-sailynoja-visual-predictive-checks/#sec-visual-predictive-checks-for-categorical-data), or cumulative event probability. 

## Residual plots
Another missing function are residual plots, `ppc_error_scatter_vs_x`, and `ppc_error_binned` work on this field, but the scatter doesn't work for discrete observations, and `ppc_error_binned` doesn't currently support covariates on the x-axis.

![Image](https://github.com/user-attachments/assets/7faaf8f1-9b60-49d1-aae9-17fb38c38979)
- The confidence bands in these need work, but at the very least adding options for binned residual plots and PAV-adjusted residuals would be an improvement.

TeemuSailynoja:
This issue could work as a general discussion on functionality for discrete distributions. I already have bayesplot-like implementations of these from the article, but would appreciate input on the best details of arguments and features.

TeemuSailynoja:
The binned residual plots are also mentioned in #263, and binned calibration plots in #150.

jgabry:
This all sounds good, thanks Teemu! Let me know which specific feedback you want. I think adding all of this is a good idea.

jgabry:
It does sound nice to be able to include the uncertainty about `p`, but I also agree that it's good to avoid too much complexity in the functions. I suppose we could add an optional argument for providing the draws of event probabilities but this argument would never be required. Is that basically what you were suggesting?

TeemuSailynoja:
Here is a quick demo of how this `ppc_calibration_overlay` would show the posterior uncertainty of the calibration.
This is a raw version without proper theming, and could for example be made into a ribbon.

On the left, pava-calibration plots for two models,
[...]

PR #352, opened 2025-05-19
PPC Calibration plots

TeemuSailynoja:
This is my work in progress of the pava calibration plots discussed in #343 

## Currently implemented:
- `ppc_calibration_overlay()`
- `ppc_calibration_overlay_grouped()`
- `ppc_calibration()`
- `ppc_calibration_grouped()`
- `ppc_loo_calibration()`
- `ppc_loo_calibration_grouped()`
- `ppc_calibration_data()`

## Needs:
- [x] Fast example to test functions.
- [x] Fix intervals in `ppc_calibration()`
   + update (2026-05-05): using currently only pointwise CI @florence-bockting
- [x] Also example use in documentation
- [x] LOO versions
- [x] Should `.ppc_calibration_data()` be exposed to users?
   + update (2026-04-15): user-facing function after refactoring by @florence-bockting 
- [x] tests
- [x] check that the input parameter names and default values make sense and are intuitive
- [x] Add documentation and comments to the code also.
   + update (2026-05-04): included a vignette that documents ppc calibration plots @florence-bockting
   + update (2026-05-06): updated function documentation @florence-bockting


TeemuSailynoja:
# Examples
These should allow for some tests of these functions.

## Creating example data
```r
library(bayesplot)
ymin <- range(example_y_data(), example_yrep_draws())[1]
ymax <- range(example_y_data(), example_yrep_draws())[2]
# Observations and posterior predictive probabilitites.
y <- rbinom(length(example_y_data()), 1, (example_y_data() - ymin) / (ymax - ymin))
prep <- (example_yrep_draws() - ymin) / (ymax - ymin)
groups <- example_group_data()
```

## PAVA Calibration overlay
Basic
```r
ppc_calibration_overlay(y, prep[1:50,])
```
image

Grouped
```r
ppc_calibration_overlay_grouped(y, prep[1:50,], groups)
```
image


## PAVA Calibration
This isn't yet quite what we want. Now the interval is not what we show in the paper. There, we use consistency intervals, that is, intervals centered at the diagonal displaying, where the calibration curve *should* lie, i.e. the posterior mean should stay within these bounds.
In this implementation, I'm plotting a confidence interval, which shows, where we think the curve lies, i.e. the diagonal should be included.

```r
ppc_calibration(y, prep)
```
image

```r
ppc_calibration_grouped(y, prep, groups)
```
image

R/ppc-calibration.R
```diff
@@ -0,0 +1,224 @@
...
+    group <- rep(1, n_obs * nrow(prep))
+  }
+
+  if (requireNamespace("monotone", quietly = TRUE)) {
+    monotone <- monotone::monotone
+  } else {
+    monotone <- function(y) {
+      stats::isoreg(y)$yf
+    }
+  }
```
jgabry: Is there an advantage to using `monotone::monotone` instead of `stats::isoreg`?

R/ppc-calibration.R
```diff
@@ -0,0 +1,224 @@
...
+#' prep <- (example_yrep_draws() - ymin) / (ymax - ymin)
+#'
+#' ppc_calibration_overlay(y, prep[1:50, ])
+NULL
+
+
+#' @rdname PPC-calibration
+#' @export
+ppc_calibration_overlay <- function(
+    y, prep, ..., linewidth = 0.25, alpha = 0.5) {
```
jgabry: So for these functions `prep` is a matrix of probabilities and not actually a matrix of draws of binary outcomes from the posterior predictive distribution, right? I think in that case the argument name `prep` makes sense. But the description at the top of the file says

> Assess the calibration of the predictive distributions `yrep` in relation to the data `y'

which makes it sound like the user should give us `yrep`. So I think we just need to reconcile how we describe this to the user.

jgabry (commented):
This all sounds good, thanks @TeemuSailynoja. I made a few small review comments/questions. In addition to those questions, when you say 

> This isn't yet quite what we want. Now the interval is not what we show in the paper. 

you mean that we will want to change this to use the consistency intervals you use in the paper, right? Do you think it's at all useful to give the user the option to choose which kind of interval? Or just strictly better to use the consistency intervals? I hadn't really thought about that.

R/ppc-calibration.R
jgabry: That is, does it do something slightly better? Or the same thing more efficiently? I've seen `stats::isoreg` before but I had never seen the monotone package. If there's no difference then it's probably not worth checking for the monotone package. If it's better then we could put monotone in Suggests and then check for it like you do here.

TeemuSailynoja: `monotone` offers an implementation of the algorithm that is noticeably faster for large samples.
I think it would be good to add it to the suggests.

R/ppc-calibration.R
TeemuSailynoja: Your interpretation is right. The description needs more clarity. `yrep` can be made to be accepted by `ppc_calibration` (without overlay).

TeemuSailynoja:
> This all sounds good, thanks @TeemuSailynoja. I made a few small review comments/questions. In addition to those questions, when you say
> 
> > This isn't yet quite what we want. Now the interval is not what we show in the paper.
> 
> you mean that we will want to change this to use the consistency intervals you use in the paper, right? Do you think it's at all useful to give the user the option to choose which kind of interval? Or just strictly better to use the consistency intervals? I hadn't really thought about that.

Leaving an option to choose is perhaps the best, as long as the difference is explained in the documentation.

Confidence = "Where do we think the calibration curve for our model lies."
Consistency = "Where should the curve of a consistent model lie."

R/ppc-calibration.R
jgabry: Ok sounds good

jgabry:
Ok great, thanks for the replies. Sounds good to me.

florence-bockting:
I've now updated the plotting functions (`pcc_calibration*`).
Unfortunately, I wasn't able to use the most recent changes from this PR; the code was failing because of several undefined arguments in the functions. So, I went one commit backwards in history and rebuilt the "confidence" vs. "consistency" version from scratch. At the same time, I created a short vignette that explains the logic behind it and shows the current visualizations together with their customization options (for convenience you find a rendered html attached).
The simulated data used in the Vignette isn’t ideal yet (I’m still building some intuition for how to best construct the dataset), so I would appreciate any input or suggestions you might have.

Please take a look at the changes whenever you have a moment @avehtari, @TeemuSailynoja. I'm happy to receive any feedback and improve things further based on your thoughts. Thank you!

[ppc-calibration.html](https://github.com/user-attachments/files/26602733/ppc-calibration.html)

avehtari:
The first comment says
> Currently implemented:
ppc_calibration_overlay()
ppc_calibration_overlay_grouped()
ppc_calibration()
ppc_calibration_grouped()
.ppc_calibration_data() - internal function

and then
> Needs:
> ...
>  - [x] LOO versions

It would be nice to list the loo versions in the list of implemented functions.

We need also better documentation for what are the expected arguments and how the plots are computed. With `reliabilittydiag()` I had used 
```
rd <- reliabilitydiag(
  EMOS = pmin(E_loo(0 + (posterior_predict(fit_betabinomial2b) > th), 
                    loo(fit_betabinomial2b)$psis_object)$value, 1),
  y = as.numeric(cu_df_b$cu > th)
)
```
where posterior_predict gives posterior draws of counts and not probabilities, and then E_loo compute LOO-expectation so that we get single probability for each observation and `reliabilitydiag` uses then bootstrap idea for the variability. Now `ppc_calibration_overlay()` / `ppc_calibration()` assumes we directly have probabilities (which we soon can get easily from brms).

In addition, for `ppc_loo_calibration()` it would be useful to mention that resampling is used, as some this is different from use of `E_loo()`

florence-bockting:
> It would be nice to list the loo versions in the list of implemented functions.

done

florence-bockting:
> We need also better documentation for what are the expected arguments and how the plots are computed.

See corresponding vignette: vignettes/ppc-calibration.Rmd

florence-bockting:
@jgabry @avehtari This repo is ready for final review. It would be nice if you can have a look. 
I created also a vignette for the ppc-calibration functions. This vignette should be "online only" as it introduces brms as dependency which we probably don't want to have as dependency in bayesplot "just" because of the vignette.

Probably you can pay specific attention to how I incorporated this vignette at the moment as I was not sure how to do this correctly. 

Thank you!

R/ppc-calibration.R
```diff
@@ -0,0 +1,719 @@
...
+    if (any(prep > 1 | prep < 0)) {
+      stop("Values of 'prep' should be predictive probabilities between 0 and 1.")
+    }
+    d <- .ppd_data(prep, group = group) |>
+      group_by(.data$group, .data$rep_id) |>
+      mutate(
+        ord = order(.data$value),
+        y_id = .data$ord,
+        value = .data$value[.data$ord],
+        cep = monotone(y[.data$ord])
```
jgabry: I think here: 

- ord = order(.data$value) is position-within-group
- cep = monotone(y[.data$ord]) indexes the full y

For any group after the first, does this compute CEPs using the wrong observations? Do we need to preserve the original y_id from .ppd_data() and index y with the ordered original IDs?

R/ppc-calibration.R
```diff
@@ -0,0 +1,719 @@
...
+ppc_calibration <- function(
+    y,
+    prep = NULL,
+    yrep = NULL,
+    prob = .95,
+    interval = c("confidence", "consistency"),
+    help_text = TRUE,
+    B = 200,
+    show_mean = TRUE,
+    show_qdots = TRUE,
```
jgabry, line 211: If show_qdots = TRUE by default then the ggdist package is _required_ for the default plot. But we only list ggdist in Suggests. So maybe we should either move ggdist to Imports or make the default show_qdots = FALSE. If we do move ggdist to Imports than any other functions in the package that use it can remove code that checks for it (code like `suggested_package("ggdist")`, which we have later in this file and in other files). What do you think?

R/ppc-calibration.R
```diff
@@ -0,0 +1,719 @@
...
+  } else {
+    w
+  }
+  total <- sum(p)
+  if (!is.finite(total) || total <= 0) {
+    rep(1 / length(w), length(w))
+  } else {
+    p / total
+  }
+}
```
jgabry: If I understand correctly, this function treats weights as log weights only if any entry is negative. But lw is documented as log weights, and I think valid unnormalized log weights can all be positive, right? Are we always assuming already normalized log weights?

R/ppc-calibration.R
```diff
@@ -0,0 +1,719 @@
...
+  y <- validate_y(y)
+  n_obs <- length(y)
+  if (!is.null(group)) {
+    group <- validate_group(group, n_obs)
+  } else {
+    group <- rep(1, n_obs)
+  }
+  if (any(y < 0 | y > 1)) {
+    abort("'y' must contain values in [0, 1] for calibration.")
+  }
```
jgabry: This would still allow fractional y not just binary y. Do we want to check if all y are equal to 0 or 1 or are we allowing e.g. y = 0.1?

R/ppc-calibration.R
```diff
@@ -0,0 +1,719 @@
...
+    if (!is.null(prep)) {
+      prep <- validate_predictions(prep, n_obs)
+      if (any(prep < 0 | prep > 1)) {
+        abort("Values of 'prep' should be predictive probabilities in [0, 1].")
+      }
+    } else {
+      yrep <- validate_predictions(yrep, n_obs)
+      if (any(yrep < 0 | yrep > 1)) {
+        abort("Values of 'yrep' should be binary outcomes in [0, 1].")
+      }
```
jgabry: Same thing as with `y` above, this allows for `y_rep` between 0 and 1, not just equal to 0 and 1.

tests/testthat/test-ppc-calibration.R
```diff
@@ -0,0 +1,477 @@
...
+  )
+
+  expect_gg(p)
+  expect_named(p$data, c("group", "y_id", "value", "cep", "lb", "ub"))
+  expect_equal(nrow(p$data), length(y))
+  expect_lte(max(p$data$ub), 1)
+  expect_gte(min(p$data$lb), 0)
+})
+
+test_that("ppc_calibration recovers identity trend for calibrated data", {
```
jgabry, line 498: This test has `diffr::expect_doppelganger` so it needs the same `skip` logic as the other visual tests: 
```
  testthat::skip_on_cran()
  testthat::skip_if_not_installed("vdiffr")
  skip_on_r_oldrel()
```

skipping on `oldrel` is used because sometimes in the past SVGs are slightly different (not in a way that the human eye can detect) and cause failures on GitHub Actions.

R/ppc-calibration.R
```diff
@@ -0,0 +1,719 @@
...
+#' # Compare confidence vs consistency intervals
+#' ppc_calibration(y, prep, interval = "confidence")
+#' ppc_calibration(y, prep, interval = "consistency")
+NULL
+
+
+#' @rdname PPC-calibration
+#' @export
+ppc_calibration_overlay <- function(
+    y, prep, ..., prob = NULL, linewidth = 0.25, alpha = 0.2) {
```
jgabry: It seems like `prob` is ignored, right? I think we should either remove it or document that it is accepted only for API consistency but has no effect. (Unless I'm wrong and it's actually serving a purpose here)

R/ppc-calibration.R
```diff
@@ -0,0 +1,719 @@
...
+    coord_equal(xlim = params$xlim, ylim = c(0, 1), expand = FALSE) +
+    xlab("predicted probability") +
+    ylab("conditional event probability") +
+    NULL
+}
+
+#' @rdname PPC-calibration
+#' @export
+ppc_calibration_overlay_grouped <- function(
+    y, prep, group, ..., prob = NULL, linewidth = 0.25, alpha = 0.2) {
```
jgabry: Same thing about `prob` here

jgabry (changes requested):
Here's a first round of review comments, sorry for the delay!

R/ppc-calibration.R
jgabry: Actually now that I think about it, maybe we already have this problem also for other function in the package that use ggdist. Since we're now using it in multiple places in the package we could just Import it and remove all code for checking if it's installed. What do you think?

avehtari: I like ggdist and I think we could use it even more, so I'm in favor just importing it

R/ppc-calibration.R
avehtari: yes, unnormalized log weights can all be positive

R/ppc-calibration.R
florence-bockting: I agree, we can import it. 
I included **ggidst** to "imports" and removed code snippets such as `suggested_package("ggdist")` and `testthat::skip_if_not_installed("ggdist")`

R/ppc-calibration.R
florence-bockting: I updated the logic:

- I assume log weights are provided as this is also what we document as required input
- change `.loo_resampling_probs()` to `.normalize_lw` and focus only on normalization (take `exp` out of the function) 
- normalization is done on all passed `lw` (i.e., log-weights; when lw is already normalized this should simply do "nothing")

```
.normalize_lw <- function(lw) {
  if (!all(is.finite(lw))) {
    abort("All values in 'lw' must be finite.")
  }
  max_lw <- max(lw)
  log_sum_exp <- max_lw + log(sum(exp(lw - max_lw)))
  lw - log_sum_exp
}

.loo_resample_data <- function(yrep, lw, psis_object) {
  lw <- .get_lw(lw, psis_object)
  stopifnot(identical(dim(yrep), dim(lw)))

  yrep <- as.matrix(yrep)
  lw <- as.matrix(lw)

  # Resample each column (observation) with its corresponding weights.
  # Sampling indices directly is much faster than constructing draws objects.
  n_obs <- ncol(yrep)
  n_draws <- nrow(yrep)
  yrep_resampled <- matrix(NA_real_, nrow = n_draws, ncol = n_obs)

  for (i in seq_len(n_obs)) {
    probs_i <- exp(.normalize_lw(lw[, i]))
    idx_i <- sample.int(n_draws, size = n_draws, replace = TRUE, prob = probs_i)
    yrep_resampled[, i] <- yrep[idx_i, i]
  }

  # Add observation names if available
  if (!is.null(colnames(yrep))) {
    colnames(yrep_resampled) <- colnames(yrep)
  }

  yrep_resampled
}
```

R/ppc-calibration.R
florence-bockting: Thank you for spotting this. Changed.

R/ppc-calibration.R
florence-bockting: thank you for spotting this. changed.

tests/testthat/test-ppc-calibration.R
florence-bockting: thank you for the explanation. Done

R/ppc-calibration.R
florence-bockting: Yes. I removed the argument and updated the documentation.

R/ppc-calibration.R
florence-bockting: Yes. I removed the argument and updated the documentation.

R/ppc-calibration.R
florence-bockting: Yes, indeed here was a logical mistake.
I changed it now such that the original index is preserved and used to index y before applying the ordering:

```
d <- .ppd_data(prep, group = group) |>
      group_by(.data$group, .data$rep_id) |>
      mutate(
        ord = order(.data$value),
        y_id_orig = .data$y_id, # for selecting groups in y
        y_id = .data$ord,
        value = .data$value[.data$ord],
        cep = monotone(y[.data$y_id_orig[.data$ord]])
      ) |>
      ungroup() |>
      dplyr::select(dplyr::all_of(c("group", "y_id", "rep_id", "value", "cep")))
```

jgabry: This looks correct to me, thanks. What do you think about adding a test like this? 

```r
  test_that("grouped overlay CEP uses original y indices within each group", {
    y <- c(0, 0, 1, 1)
    group <- c("A", "A", "B", "B")
    prep <- matrix(c(0.2, 0.1, 0.2, 0.1), nrow = 1)
    d <- ppc_calibration_data(
      y = y,
      prep = prep,
      group = group,
      type = "overlay"
    )
    d_b <- d[d$group == "B", , drop = FALSE]
    expect_equal(d_b$value, c(0.1, 0.2))
    expect_equal(d_b$cep, c(1, 1))
  })
```

I'm pretty sure (not 100%) that this tests for the right behavior.

R/ppc-calibration.R
jgabry: Looks good to me. @avehtari do you see any issues I'm missing?

R/ppc-calibration.R
```diff
@@ -489,8 +492,11 @@ ppc_calibration_data <- function(
       }
     } else {
       yrep <- validate_predictions(yrep, n_obs)
-      if (any(yrep < 0 | yrep > 1)) {
-        abort("Values of 'yrep' should be binary outcomes in [0, 1].")
+      if (!all(yrep == 0 | yrep == 1)) {
+        abort(sprintf(
+          "Values of 'yrep' must be 0 or 1. Got range [%g, %g].",
+          min(yrep), max(yrep)
+        ))
```
jgabry: I think this error message (and the one for y above) could be potentially confusing. If the user puts y_rep = 0.5 then they could get a message saying: 

```
Values of 'yrep' must be 0 or 1. Got range [0, 1].
```

which is technically true but "range [0, 1]" is a bit strange since it doesn't give them any useful information. Maybe just the simpler message is enough? That is, just `Values of 'yrep' must be 0 or 1`.

jgabry (commented):
Looks good, thanks. Just two minor comments, one about a test we could add and one about the error message for non-binary y/y_rep.

R/ppc-calibration.R
avehtari: I think log weight -Inf is valid, that is, weight can be 0.

florence-bockting: Yes. Thanks. -Inf should work as well. I adjusted it.

R/ppc-calibration.R
florence-bockting: Alright. I adjusted it and used the shorter message without information about a "range".

R/ppc-calibration.R
florence-bockting: Looks good to me. I also added a second test.

florence-bockting:
> Looks good, thanks. Just two minor comments, one about a test we could add and one about the error message for non-binary y/y_rep.

Thank you. I have updated the code according to your comments.

R/ppc-calibration.R
```diff
@@ -0,0 +1,709 @@
...
+  # Add observation names if available
+  if (!is.null(colnames(yrep))) {
+    colnames(yrep_resampled) <- colnames(yrep)
+  }
+
+  yrep_resampled
+}
+
+.normalize_lw <- function(lw) {
+  if (any(lw == Inf)) {
```
avehtari: Maybe we should check for NaN and NA, too?

florence-bockting: Thank you. I updated the check to include also is.na().

jgabry (approved):
Thanks! I guess this is ready? @avehtari any other comments/suggestions or ready to merge?

TeemuSailynoja:
@florence-bockting thank you for taking this over the finish line! I just read through the  vignette and I think it's great. Clear and thorough.

PR #353, opened 2025-05-26
Add possibility to control extrapolation in ppc_km_overlay() and ppc_km_overlay_grouped()

### LINKED ISSUE #346: Add parameter extrapolation_factor to ppc_km_overlay() to control extrapolation
The first figure shows how the function currently behaves. The plot extrapolates unreasonably much. I propose adding a parameter extrapolation_factor with a default value of 1.2 (20 % extrapolation) to control the extrapolation. The second figure shows the plot for the same data with extrapolation_factor = 1.2. In case one needs to extrapolate further, one can easily do this by adjusting the parameter. I have implemented the change in my own [fork](https://github.com/Sakuski/bayesplot/tree/improve-ppc_km_overlay).

Alternatively, we can keep the default behavior unaltered, but I think we should still have the parameter so that we have the possibility to control extrapolation if needed.

I don't see any reason to set extrapolation_factor < 1 so in my current implementation the function fails with an error message if one tries to set extrapolation_factor < 1 as I believe this situation is most likely caused by a user misunderstanding the parameter etc.

Image

Image

Sakuski:
Fixes #346


jgabry:
I'll wait for the unit tests to run again but after that I think this is ready to merge. Thanks again.

Issue #354, opened 2025-06-11
Adding quantile dot plots to PPC distributions


behramulukir:
Right now, `bayesplot` offers various functions for [posterior predictive check](https://mc-stan.org/bayesplot/reference/PPC-distributions.html) such as `ppc_dens`, `ppc_hist`, `ppc_freqpoly`, `ppc_boxplot`, and a couple of others. One of the plots that `bayesplot` currently doesn't have a function for is quantile dot plots.

In [their preprint](https://teemusailynoja.github.io/visual-predictive-checks/),  Säilynoja et al compared KDE plots, histograms, and quantile dot plots, finding that "compared to KDEs, quantile dot plots have the added benefit of allowing for fast visual probability estimation in the tails of the distribution" and "in case of discontinuities and outliers, quantile dot plots also offer a better visualization than kernel density plots and histograms."

<img width="843" alt="Image" src="https://github.com/user-attachments/assets/25f12c4f-e4b0-4887-8f5f-0e1cf02ccec5" />

Based on that, I suggest adding quantile dot plots. It has certain benefits that are not currently offered by the package and gives an alternative visualisation method to users. 

For this implementation, I am thinking of using `geom_dotplot` function, which should be fit for the task.

jgabry:
Sounds like a good idea! My guess is that you can probably use most of the same internal code that we use for the histograms and KDE plots but then pass it to `geom_dotplot` instead.

behramulukir:
Thanks, I'll work on this now.

PR #357, opened 2025-06-20
add quantile dot plot functions

behramulukir:
As I mentioned at #354 I have been working on adding quantile dot plots. It was more complicated than I was expecting, so I wanted to make sure about the direction of this.

## Current Progress
So far, I have implemented two functions, `ppc_qdotplot` which under the hood utilises `geom_dotplot` and `ppc_qdotplot_ggdist` which utilises `stat_dots` from `ggdist`.

### `qdotplot`
This function utilises `geom_dotplot`, which gives a very simple way to build dot plots. `geom_dotplot` function has two main deficiencies:

1. Very often, the number of dots exceeds the space that the graphic has, and since `geom_dotplot` doesn't have a proper auto-sizing for dots, the visual is getting clipped and not presenting the full picture.
2. It doesn't have functionality to plot according to quantiles.

To minimise the first problem, I implemented a simple auto-sizing algorithm which performs well when the bin width is not specified or very reasonably specified; however, when it is even a little unsuitable, it is likely to run into issues. Here are some example uses of `ppc_qdotplot`: 

```r
y <- example_y_data()
yrep <- example_yrep_draws()
group <- example_group_data()
ppc_qdotplot(y, yrep[1:8, ])
```
 plot1

```r
ppc_qdotplot(y, yrep[1:8, ], binwidth = 5)
```
plot2

```r
ppc_qdotplot(y, yrep[1:8, ], binwidth = 3)
```
plot6

```r
ppc_qdotplot(y, yrep[1:8, ], binwidth = 10)
```

plot3

As you can see, for certain bin widths, dots are over-compressed or exceed the limit of the graphic.

### `qdotplot_ggdist`
The second alternative relies on `stat_dots` from  `ggdist`. It is slightly more successful overall in handling edge cases and has a warning system; however, it also brings additional dependency to the package, making me unsure if we want to go that route. The advantages of this over `qdotplot` are:

1. When dots are unfit for the graph size, it gives a warning to the user
2. It can plot based on quantiles

```r
ppc_qdotplot_ggdist(y, yrep[1:8, ])
```
plot4

```r
ppc_qdotplot_ggdist(y, yrep[1:8, ], binwidth = 3)
```
plot7

```r
ppc_qdotplot_ggdist(y, yrep[1:8, ], binwidth = 5)
```
plot5

```r
ppc_qdotplot_ggdist(y, yrep[1:8, ], binwidth = 10)
```
plot8

In the previous three graphics where the dots do not fit into the visual, the following error is thrown at the console:
```r
Warning messages:
1: The provided binwidth will cause dots to overflow the boundaries of the geometry.
→ Set `binwidth = NA` to automatically determine a binwidth that ensures dots fit within the bounds,
→ OR set `overflow = "compress"` to automatically reduce the spacing between dots to ensure the dots fit within the bounds,
→ OR set `overflow = "keep"` to allow dots to overflow the bounds of the geometry without producing a warning.
```

`ppc_qdotplot_ggdist` can be used with quantiles as follows:

```r
ppc_qdotplot_ggdist(y, yrep[1:8, ], quantile = 25)
```
plot9

## Future
To move forward, I think we need to decide on which function we want to keep and whether we want to add `ggdist` dependency. After that, I can remove the unnecessary function and continue with the rest of the implementation.

- [X] implement `geom_dotplot` version
- [X] implement `stat_dots`
- [x] decide on which function to keep
- [x] implement the ppd version
- [x] update the documentation
- [x] implement tests


jgabry:
I didn't realize there would be this sort of difficulty with the dot plots. I think handling those edge cases and having appropriate warnings is probably enough of a benefit to add ggdist dependency and maybe we can also use ggdist for other plots in the future. For now we could just add it to Suggests (instead of Imports) and then use our internal `suggested_package()` function to make sure it's installed if someone wants to make the dot plots. If we end up using ggdist for a bunch of other plots in the future we can move it to Imports so that it gets automatically installed. @avehtari @TeemuSailynoja what do you think?

behramulukir:
This solution makes sense to me. If @TeemuSailynoja or @avehtari don't have any objection, I can work on this.

avehtari:
I like ggdist, so I'm fine using it more

TeemuSailynoja:
I think adding ggdist as a suggested package is very good for this.

DESCRIPTION
```diff
@@ -55,7 +55,8 @@ Suggests:
     shinystan (>= 2.3.0),
     survival,
     testthat (>= 2.0.0),
-    vdiffr (>= 1.0.2)
+    vdiffr (>= 1.0.2),
+    ggdist
```
jgabry: Not a big deal, but can you move ggdist up in the list so that it's in alphabetical order like all the other packages listed?

R/ppc-distributions.R
```diff
@@ -507,6 +515,45 @@ ppc_boxplot <-
       xaxis_title(FALSE)
   }
 
+#' @rdname PPC-distributions
+#' @export
+#' @template args-qdotplot
+ppc_qdotplot <-
+  function(y,
+           yrep,
+           ...,
```
jgabry, line 524: In the documentation for `ggdist::stat_dots` it looks like there are various arguments that can be passed via `...` like `dotsize`, `stackratio`, `layout`, and some others. Should we pass `...` to `stat_dots` so users can specify these arguments if they want? Is there a reason for us not to allow that?

man-roxygen/args-qdotplot.R
```diff
@@ -0,0 +1,3 @@
+#' @param quantiles For dot plots, optional integer `quantiles` defines the
+#' number of quantiles to plot in dot plot to produce a quantile dot plot.
+#' If `quantiles` is `NA` (the default) then all data points are plotted.
```
jgabry:
```suggestion
#' @param quantiles For dot plots, an optional integer passed to
#'   [ggdist::stat_dots()] specifying the number of quantiles to use for a
#'   quantile dot plot. If `quantiles` is `NA` (the default) then all data
#'   points are plotted.
```

tests/testthat/test-ppc-distributions.R
```diff
@@ -58,6 +58,22 @@ test_that("ppc_dens,pp_hist,ppc_freqpoly,ppc_boxplot return ggplot objects", {
...
+  expect_gg(ppc_qdotplot(y2, yrep2, binwidth = 0.1, quantiles = 25))
+
+
+  # ppd versions
+  expect_gg(ppd_qdotplot(y, yrep[1:8, ]))
+  expect_gg(ppd_qdotplot(yrep[1,, drop = FALSE], quantiles = 25))
+  expect_gg(ppd_qdotplot(yrep[1:8, ], binwidth = 0.1))
+  expect_gg(ppd_qdotplot(yrep2, binwidth = 0.1, quantiles = 25))
+})
+
```
jgabry: If some (or all) of these will have warning messages about overflow, can you wrap the relevant ones in `expect_warning()`?

tests/testthat/test-ppc-distributions.R
```diff
@@ -177,6 +193,38 @@ test_that("ppc_boxplot renders correctly", {
...
+  p_binwidth <- ppd_qdotplot(vdiff_yrep[1:8, ], binwidth = 3)
+  vdiffr::expect_doppelganger("ppd_qdotplot (binwidth)", p_binwidth)
+
+  p_quantile <- ppd_qdotplot(vdiff_yrep[1:8, ], quantiles = 50)
+  vdiffr::expect_doppelganger("ppd_qdotplot (quantile)", p_quantile)
+
+  p_quantile_binwidth <- ppd_qdotplot(vdiff_yrep[1:8, ], binwidth = 3, quantiles = 50)
+  vdiffr::expect_doppelganger("ppd_qdotplot (quantile-binwidth)", p_quantile_binwidth)
+})
+
```
jgabry: Same thing as above about warning messages. Aside from checking that they throw a warning when we expect them too, the logs test output also gets very messy when we have a lot of warnings. Wrapping them is expect_warning() will get rid of that.

jgabry (changes requested):
Thanks for this! I still need to actually play around with the functions and try using them to make some plots (I will try to do that soon), but the code is looking good. I made a few small review comments and, in addition to those, two other things: 

1. It looks like several tests are failing. Can you look into why they're failing?  

2. Can you regenerate the documentation using `devtools::document()`. That will regenerate the user manual pages (the .Rd files in the `man` directory) based on the documentation you put in the R files. 

3. Thinking about the name for this function, maybe we should just call it `ppc_dots()`. What do others think?

behramulukir:
Thanks, I'll look into all of these and update the code

R/ppc-distributions.R
behramulukir: No, we can definitely allow more arguments. I was unsure about this but decided to keep options limited to be safe. I'll add those arguments now

jgabry: For this I recommend just letting the user pass them to `stat_dots()` via `...` instead of adding a named argument for each of them (at least if you're going to allow them to pass a lot of arguments, if just a couple you could add them individually). To do this you need to do the following: 

- pass `...` to `stat_dots()`
- change the doc for `...` from `#' @param ... Currently unused.` to `#' @param ... For dot plots, optional additional arguments to pass to [ggdist::stat_dots()]`. 
- change the line `check_ignored_arguments(...)` to `check_ignored_arguments(..., ok_args = c("dotsize", "layout"))` but also add the other arguments that seem potentially useful in addition to `dotsize` and `layout` to `ok_args`. `ok_args` just tells bayesplot not to throw a warning if those arguments are passed via `...`

I think that's it. Let me know if you have any questions about that.

behramulukir: Sure, thank you so much

Issue #360, opened 2025-07-21
Emphasizing discreteness of data in rootograms


behramulukir:
In [their preprint](https://teemusailynoja.github.io/visual-predictive-checks/), Säilynoja et al proposed a new look for rootograms that emphasises the discreteness of the data more compared to the version of `ppc_rootogram` that is currently used in bayesplot. Below, you see an example of the current version of rootogram and the version with proposed changes.

**Current version:**
![Current ppc_rootogram](https://teemusailynoja.github.io/visual-predictive-checks/images/03_01_rootogram_plain_cut-2.png)
**Proposed version:**
![Proposed ppc_rootogram](https://teemusailynoja.github.io/visual-predictive-checks/images/03_01_rootogram_discrete_cut-1.png)

There are two main differences between the current version and the proposed version. First, the proposed version shows predictive frequencies as points and point-wise credible intervals, instead of connected lines and filled areas. This change highlights the discreteness of predictions compared to the current version, which is not explicit about this. Second change is about how square root scaling is implemented on the vertical axis. Currently, the axis is labelled by equal tick marks of square roots of the frequencies, but what is proposed instead is transforming the axis to a square root scale and displaying the untransformed frequencies, which can be seen at the proposed visual.

Both of these changes make sense to me, and to not break graphs that people are already working on, we can implement them connected to a parameter such as `discrete`, which would be `false` by default and when it is set to `true`, it creates rootograms with new, proposed aesthetics. Of course, we can add a warning now to let users know about these changes and deprecate the old version after several releases.

PR #362, opened 2025-07-30
adding discrete option to ppc_rootogram

behramulukir:
In #360, I explained the rationale behind adding a new `discrete` option to the `style` argument of `ppc_rootogram()` to offer a new plot aesthetic which puts more emphasis on the discreteness of the data. This PR is about my work on that issue. The only difference between my implementation and what @TeemuSailynoja proposed is that I decided to use rhombuses instead of dots for `Observed` values for better distinction.

This implementation passes all existing tests, so I think the current style options for `ppc_rootogram()` remain unchanged.

- [x] Add `discrete` option to `style` argument
- [x] Implement new tests
- [x] Update the documentation

## Examples
Here are some examples of how `ppc_rootogram()` works with `discrete` style.

```r
y <- rpois(100, 20)
yrep <- matrix(rpois(10000, 20), ncol = 100)
ppc_rootogram(y, yrep, style = "discrete")
``` 
rootogram1

```r
ppc_rootogram(y, yrep, style = "discrete", prob=0.7)
``` 
rootogram2

```r
ppc_rootogram(y, yrep, style = "discrete", size=2)
``` 
rootogram3

### DIFF
--- a/R/ppc-discrete.R
+++ b/R/ppc-discrete.R
@@ -21,9 +21,12 @@
 #' @param size,fatten,linewidth For bar plots, `size`, `fatten`, and `linewidth`
 #'   are passed to [ggplot2::geom_pointrange()] to control the appearance of the
 #'   `yrep` points and intervals. For rootograms `size` is passed to
-#'   [ggplot2::geom_line()].
+#'   [ggplot2::geom_line()] and [ggplot2::geom_pointrange()].
 #' @param freq For bar plots only, if `TRUE` (the default) the y-axis will
 #'   display counts. Setting `freq=FALSE` will put proportions on the y-axis.
+#' @param bound_distinct For `ppc_rootogram(style = "discrete)`,
+#'  if `TRUE` then the observed counts will be plotted with different shapes
+#'  depending on whether they are within the bounds of the expected quantiles.
 #'
 #' @template return-ggplot-or-data
 #'
@@ -52,10 +55,12 @@
 #'   style can be adjusted to focus on different aspects of the data:
 #'   * _Standing_: basic histogram of observed counts with curve
 #'    showing expected counts.
-#'   * _Hanging_: observed counts counts hanging from the curve
+#'   * _Hanging_: observed counts hanging from the curve
 #'    representing expected counts.
 #'   * _Suspended_: histogram of the differences between expected and
 #'    observed counts.
+#'    * _Discrete_: a dot-and-whisker plot of the expected counts and dots
+#'   representing observed counts
 #'
 #'   **All of the rootograms are plotted on the square root scale**. See Kleiber
 #'   and Zeileis (2016) for advice on interpreting rootograms and selecting
@@ -213,7 +218,7 @@ ppc_bars_grouped <-
 #' @rdname PPC-discrete
 #' @export
 #' @param style For `ppc_rootogram`, a string specifying the rootogram
-#'   style. The options are `"standing"`, `"hanging"`, and
+#'   style. The options are `"discrete", "standing"`, `"hanging"`, and
 #'   `"suspended"`. See the **Plot Descriptions** section, below, for
 #'   details on the different styles.
 #'
@@ -234,13 +239,15 @@ ppc_bars_grouped <-
 #'
 #' ppc_rootogram(y, yrep, style = "hanging", prob = 0.8)
 #' ppc_rootogram(y, yrep, style = "suspended")
+#' ppc_rootogram(y, yrep, style = "discrete")
 #'
 ppc_rootogram <- function(y,
                           yrep,
                           style = c("standing", "hanging", "suspended", "discrete"),
                           ...,
                           prob = 0.9,
-                          size = 1) {
+                          size = 1,
+                          bound_distinct = TRUE) {
   check_ignored_arguments(...)
   style <- match.arg(style)
   y <- validate_y(y)
@@ -268,7 +275,7 @@ ppc_rootogram <- function(y,
   tyrep[is.na(tyrep)] <- 0
 
   #Discrete style
-  pred_mean <- colMeans(tyrep)
+  pred_median <- apply(tyrep, 2, median)
   pred_quantile <- t(apply(tyrep, 2, quantile, probs = probs))
   colnames(pred_quantile) <- c("lower", "upper")
 
@@ -278,28 +285,40 @@ ppc_rootogram <- function(y,
   y_count[is.na(y_count)] <- 0
 
   if (style == "discrete") {
-    obs_shape <- ifelse(y_count >= pred_quantile[, "lower"] & y_count <= pred_quantile[, "upper"], 24, 25)
+    if (bound_distinct) {
+      # If the observed count is within the bounds of the predicted quantiles,
+      # use a different shape for the point
+      obs_shape <- obs_shape <- ifelse(y_count >= pred_quantile[, "lower"] & y_count <= pred_quantile[, "upper"], "In", "Out")
+    } else {
+      obs_shape <- rep("Observed", length(y_count)) # all points are the same shape for obsved
+    }
 
     data <- data.frame(
       xpos = xpos,
       obs = y_count,
-      pred_mean = pred_mean,
+      pred_median = pred_median,
       lower = pred_quantile[, "lower"],
       upper = pred_quantile[, "upper"],
       obs_shape = obs_shape
     )
     # Create the graph
     graph <- ggplot(data, aes(x = xpos)) +
-      geom_pointrange(aes(y = pred_mean, ymin = lower, ymax = upper, color = "Expected"), fill = get_color("d"), linewidth = size, size = size, fatten = 2, alpha = 0.65) +
-      geom_point(aes(y = obs, shape=ifelse(obs_shape==24, "In", "Out")), size = size * 2, color = get_color("lh"), fill = get_color("lh")) +
+      geom_pointrange(aes(y = pred_median, ymin = lower, ymax = upper, color = "Expected"), fill = get_color("lh"), linewidth = size, size = size, fatten = 2, alpha = 1) +
+      geom_point(aes(y = obs, shape = obs_shape), size = size * 1.5, color = get_color("d"), fill = get_color("d")) +
       scale_y_sqrt() +
-      scale_fill_manual("", values = get_color("lh"), guide="none") +
-      scale_color_manual("", values = get_color("dh")) +
+      scale_fill_manual("", values = get_color("d"), guide="none") +
+      scale_color_manual("", values = get_color("lh")) +
       labs(x = expression(italic(y)), y = "Count") +
       bayesplot_theme_get() +
       reduce_legend_spacing(0.25) +
-      scale_shape_manual(values = c("Out"=24, "In"=25), guide = "legend") +
-      guides(shape = guide_legend(" Observation \n within bounds"))
+      scale_shape_manual(values = c("In" = 22, "Out" = 23, "Observed" = 22), guide = "legend")
+      if (bound_distinct) {
+        graph <- graph +
+          guides(shape = guide_legend(" Observation \n within bounds"))
+      } else {
+        graph <- graph +
+          guides(shape = guide_legend(""))
+      }
     return(graph)
   }
 
--- a/R/ppc-discrete.R
+++ b/R/ppc-discrete.R
@@ -256,103 +256,23 @@ ppc_rootogram <- function(y,
                           bound_distinct = TRUE) {
   check_ignored_arguments(...)
   style <- match.arg(style)
-  y <- validate_y(y)
-  yrep <- validate_predictions(yrep, length(y))
-  if (!all_counts(y)) {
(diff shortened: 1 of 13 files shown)


jgabry:
Thanks @behramulukir. I like this new style. One question for now: should we give users control over the color of the observed points since these colors may not be optimal for all the available color schemes? 

In most of the other plots the colors for everything come from the color scheme, but I see that it would be difficult to do that here if we want to really distinguish between observations that fall inside and outside the interval. We don't usually have arguments that control the colors, but I guess we could do that here. 

Another option would be to use the shape of the point (instead of the color) to distinguish whether it falls inside and outside the interval and then use a color that comes from the chosen color scheme.

behramulukir:
Yes @jgabry, I think it would be better to offer more colour options, which would be more inclusive. The current colour choices may be indistinguishable for some types of colour blindness. The only issue here is not confusing users about that colour control arguments can only be used with `discrete` style and not other styles. This can be achieved by correct documentation, but I need to think about it.

Distinguishing by the shape of the point also makes sense, but then we need to include both the inside interval and outside interval shapes in the legend to make things clear. 

I'll think about which implementation is less confusing for the users and try out different implementations.

behramulukir:
Based on our discussions with @TeemuSailynoja, we decided that it is a better choice to indicate observations within and out of the bounds by shapes of the points instead of colours, because as you said those colors should be available for each color scheme, however, adding a dedicated color argument that would only work if the `style = "discrete"` would be confusing for users in addition to the fact that bayesplot doesn't generally have a such argument. Instead, having determined shapes for observations within and out of the bounds and following the colour scheme chosen by the user for all points and lines seemed more logical.

However, there are nuances to these changes which we need to decide on before merging this PR to main.

## Shapes
As we decided to indicate whether observed points fall into intervals or not, the first shape option that we picked was triangles. So, we used downward triangles for observation points that fall into intervals and upward triangles for observation points that fall outside of the intervals. Here is an example of it:

discrete_triangle

We also had the idea of using squares to indicate that observed points are within the intervals and rhombuses to indicate they are not. Here is an example of it:

discrete_square

We didn't settle on any of these options, so please tell me which one you think is better looking.

## Colors
Another point of building this new style of rootograms was deciding on how to colour different elements. Since we decided not to use colour to indicate the position of observed points, all of the visual elements should follow from the colour scheme that users pick. To do that, there were two main paths: following from other [rootogram styles (standing, hanging, suspended)](https://mc-stan.org/bayesplot/reference/PPC-discrete.html), where observed points are light and expected points are dark or following from [PPC intervals](https://mc-stan.org/bayesplot/reference/PPC-intervals.html), where observed points are dark and expected points are light.

The argument for following other rootogram styles, where observed points are light and expected points are dark, is that we have the same colouring practice for all kinds of rootograms. On the other hand, it is worth noting that `discrete` style is fairly different from other styles because other styles has bars for observed points which occupy considerable amount of the graph space and having them darker would make the graph visually heavy. Since this is not a case, we don't need to follow from other rootograms and instead follow `ppc_intervals` which has darker color for observed points and lighter color for expected points. The argument for this is `ppc_intervals` and `ppc_rootogram(style="discrete")` has more in common visually as both functions utilise `geom_pointrange()`.

Here is the example for lighter observed - darker expected:

discrete_colornormal

Here is the example for darker observed - lighter expected:

discrete_colorreversed

One thing we also decided to change was removing the change in alpha of the lines where until the halfway, line had higher alpha and from halfway to the end of line it was lower alpha. This wasn't contributing much to user's understanding of the situation and adding another visual effect to a graph that already has lots of different visual cues. Removing this and having uniform alpha for the full lines made the graph easier to understand.

jgabry:
Thanks for the detailed summary. Here are a few comments: 

1. Shape: I don't have a strong preference between the triangles and the squares/rhombuses. Both seem good. If anyone else (you, Teemu, Aki, TJ, etc.) has a preference we can go with that. 

2. Color: I think your point about the similarity to `ppc_intervals()` is a good one. I would vote for copying `ppc_intervals()` (darker observed - lighter expected) instead of the other rootogram styles. Also, since we might want to eventually deprecate the other rootogram styles (right?), it makes less sense to copy them. 

3. Alpha: that change seems good to me.

behramulukir:
Thanks for your comments, @jgabry! 

I have a slight preference towards using squares and rhombuses instead of triangles, since the directions of triangles don't have a meaning on their own -nothing would change if we swap the directions- but it might signal users that there actually is a meaning, so if we use squares and rhombuses, we can avoid that. So, I'll go with them instead -unless anyone has an argument against it.

For the colour, I agree with your comment. I am also in favour of copying `ppc_intervals()` because of consistency, and also because it is easier to distinguish that way. Observed points are plotted on top of expected points, and it is easier to distinguish overlapping points in a darker foreground and a lighter background compared to a lighter foreground and a darker background. And yes, I think it makes sense to slowly depreciate other rootogram styles, so I can add a depreciation warning to my code in the next commit about this.

aloctavodia:
I have a slight preference for circles (in), triangles (out). I think those are visually easier to differentiate.  And having an option to enable/disable the distinction. 
Another nitpick, the circle for the expected values should be on top of the line (as in plot_ppc_intervals), not behind (or it should be solid, alpha=1). Additionally, having the edge and face of the circle to be the same color as the line contributes to a plot that looks less cluttered.

behramulukir:
To be honest, I am in favour of using circles only for expected points, as the colours we use for distinction are not completely different colours but shades of the same colour, which might be harder to distinguish for some people, and especially for people with visual impairments. I also think this can be a future update to `ppc_intervals()` to enhance its visuality.

I agree with your comment that the expected points should be on top of their lines. I checked `ppc_intervals()`, so my guess is in our case, having `alpha = 1` would solve the issue, but anyways, I'll look into this point in detail.

behramulukir:
Based on our discussions, I made a few updates to `ppc_rootogram(style = "discrete")`.

First, I realised a critical point that apparently, in other styles of `ppc_rootogram()`, `tyexp <- sqrt(colMeans(tyrep))` is used to calculate expected counts. This doesn't create any problem with other styles, and I have been using a similar structure for the discrete style as well. However, I realised that doing this in the discrete style caused issues -especially with lower `prob` values- such as observed points that are out of the bounds being marked as within the bounds. This is because in all styles, these bounds are calculated based on quantiles. This created a weird dynamic in the discrete style that we were using the mean points as centre points of our `geom_pointrange()` while the lower and upper ends of it were based on quantiles. I decided that since the median is the actual centre of quantiles, it makes more sense to use them as centre points of `geom_pointrange()`. Here is an example of that weird case:

```r
# if mean is used as center
pred_mean <- apply(tyrep, 2, mean)
pred_quantile <- t(apply(tyrep, 2, quantile, probs = probs))
colnames(pred_quantile) <- c("lower", "upper")
geom_pointrange(aes(y = pred_mean, ymin = lower, ymax = upper, color = "Expected") ...)


ppc_rootogram(y, yrep, style = "discrete", prob=0)
``` 
disceret_mean

In the graph above, for the 7th point, both upper and lower quantiles are calculated as 0, while the mean is 0.04 and the observed count is 0. Therefore, the observed count is technically within the bounds, even though there are no visual bounds, and it doesn't overlap with the centre point.

Here is the same case with my fix:

```r
# if median is used as center
pred_median <- apply(tyrep, 2, median)
pred_quantile <- t(apply(tyrep, 2, quantile, probs = probs))
colnames(pred_quantile) <- c("lower", "upper")
geom_pointrange(aes(y = pred_median, ymin = lower, ymax = upper, color = "Expected") ...)


ppc_rootogram(y, yrep, style = "discrete", prob=0)
``` 
discrete_median

Here, as you can see, the same as the previous graph, there are no visual bounds. However, by using the median, we made sure that observed counts are marked only if they overlap with the centre of expected counts. 

I am open to suggestions about what to do in this case. Should I use mean or median? What can I do to avoid these non-optimal scenarios?

There are some visual changes as well. As suggested by @aloctavodia, now the `alpha` for `geom_pointrange()` is 1, so that circles for expected values are not behind the line, but instead now they look like one solid object. In addition to that, I added a new argument called `bound_distinct`, which is used to indicate if you want observed points to be plotted with different shapes depending on whether they are within or out of the expected bounds. When it is true (as it is by default), the observed points within the bounds are marked with squares and points out of the bounds are marked with rhombuses. When it is false, all observed points are marked with squares. Changing the `bound_distinct` also changes the legend slightly.

I also made a small aesthetic change to avoid visual blocking when observed and expected points overlap. Now, the size of observed points is slightly smaller, so when they overlap with expected points, they don't completely block the view. Here is two examples of those visual changes (using median):

```r
ppc_rootogram(y, yrep, style = "discrete")
``` 
discrete_distinct

```r
ppc_rootogram(y, yrep, style = "discrete", bound_distinct = FALSE)
``` 
discrete_nondistinct

TeemuSailynoja:
> In the graph above, for the 7th point, both upper and lower quantiles are calculated as 0, while the mean is 0.04 and the observed count is 0. Therefore, the observed count is technically within the bounds, even though there are no visual bounds, and it doesn't overlap with the centre point.

> I am open to suggestions about what to do in this case. Should I use mean or median? What can I do to avoid these non-optimal scenarios?

Having the posterior expectation outside the intervals looks weird, but if we change the plot to show medians, is that an unexpected behaviour for users?

behramulukir:
I have made some updates to the documentation and changed the labelling of the plot a bit for the `discrete` style. Here are two examples of it:

discrete
discrete_non-distinct

To make sure that the choice of median instead of mean to determine the centre points of `y`<sub>`rep` </sub> is clear, I decided to avoid using terms such as `observed` and `expected`, especially because `expected` can imply `mean` in some contexts. Instead, now the plot uses `y` and `y`<sub>`rep` </sub>, which are more neutral terms. Also now, the documentation explicitly states that `discrete` style uses median as the centre points of `y`<sub>`rep` </sub>.

As we have discussed, I haven't touched other styles to avoid changing people's existing plots and also because it is planned that those styles will be deprecated in the future, so changing their behaviour at this point doesn't make much sense.

R/ppc-discrete.R
```diff
@@ -266,6 +279,57 @@ ppc_rootogram <- function(y,
   }
   tyrep <- do.call(rbind, tyrep)
   tyrep[is.na(tyrep)] <- 0
+
+  #Discrete style
```
TeemuSailynoja: @jgabry this is at the same time a call for your opinion.
Now the function is housing two very different functions inside it, the traditional three rootograms, and the discrete one. The discrete one has a separate logical branch that has its own return statement.

They differ in
1. Data generation (slightly) mean vs. median, sqrt(counts).
2. Geoms (lines and ribbons vs points and intervals)
3. Scales (colouring and y-axis scaling)

For future maintainability, I think we could split this internally to 
- `.ppc_rootogram_data()` that handles the logic for data generation.
- define `geom_y <- ifelse ....` and `geom_yrep <- ifelse ....` and have one logical branch for constructing the final graph before returning it.

jgabry: Yeah, good suggestion, I agree it should be simplified

R/ppc-discrete.R
```diff
@@ -298,34 +282,69 @@ ppc_rootogram <- function(y,
...
+  }
+
+  # Creating the graph
+  graph <- ggplot(data)
+
+  if (style == "discrete") {
+    graph <- graph +
+      geom_yrep +
+      geom_y +
+      aes(x = xpos) +
```
jgabry: This line is causing a NOTE in R cmd check (which doesn't cause it to fail on GitHub Actions but will be noticed by CRAN): 

https://github.com/stan-dev/bayesplot/actions/runs/17111544388/job/48533894818?pr=362#step:6:111

I think it probably needs to be `.data$xpos`

tests/testthat/test-ppc-discrete.R
```diff
@@ -176,5 +177,30 @@ test_that("ppc_rootogram renders correctly", {
...
+    y = vdiff_y2,
+    yrep = vdiff_yrep2,
+    prob = 0.8,
+    size = 1,
+    style = "discrete",
+    bound_distinct = FALSE
+  )
+
+  vdiffr::expect_doppelganger(
+    title = "ppc_rootogram (style='discrete', prob, size, bound_distinct=FALSE)",
```
jgabry: We're getting a NOTE from R cmd check that the file path to the SVG that results from this is too long: 

https://github.com/stan-dev/bayesplot/actions/runs/17111544388/job/48533894818?pr=362#step:6:76

Maybe removing `style=` will make it short enough? E.g. `ppc_rootogram('discrete',...`

jgabry (changes requested):
This is looking good. @TeemuSailynoja is this what you had in mind? I made a just few minor comments due to some NOTES from R cmd check

TeemuSailynoja (approved):
I think the flow is clearer now. On my last scan through the code, I thought there was more in common between the two functions, but as even the order of the yrep_geom and y_geom is different in the two approaches, this still remained quite separated. I do like that the dataset is created in a separate function now.

jgabry (approved):

Issue #330, opened 2024-10-14
`ppc_stat` for discrete test statistics?


santikka:
It seems that `ppc_stat` always assumes that the test statistic defined by `stat` is continuous, and thus draws a histogram, which is fine but sometimes looks quite ugly, for example:
``` r
set.seed(0)
y <- rbinom(20, 1, 0.2)
yrep <- matrix(rbinom(2000, 1, prob = 0.4), 1000, 20, byrow = TRUE)
bayesplot::ppc_stat(y, yrep, function(x) sum(abs(diff(x))))
#> `stat_bin()` using `bins = 30`. Pick better value with `binwidth`.
```

![](https://i.imgur.com/gczn8N6.png)

<sup>Created on 2024-10-14 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

Is there an option (or another function) to specify that the test statistic is discrete, so that `ppc_stat` would produce a barplot instead?

jgabry:
Yeah you're right that it would look nicer as a bar plot when the test statistic is discrete. There's not a way to tell `ppc_stat` to do that right now, but I think it's a good idea. If you or anyone else is interested in adding it I would be happy to review a pull request. Otherwise I can certainly do it myself when I have time, but I'm not sure when that will be at the moment.

jgabry:
@santikka this should now be available on the master branch thanks to the recent PR from @behramulukir.

PR #369, opened 2025-08-28
adding `discrete` argument to `stat` functions

behramulukir:
As stated by @santikka at #330, currently, `stat` functions in bayesplot, such as `ppc_stat` and `ppd_stat`, completely rely on histograms while plotting. This is problematic since histograms are not optimal for discrete data, which the user might be working with. To overcome this issue, this PR adds a `discrete` argument to four `stat` functions: `ppc_stat`, `ppc_stat_grouped`, `ppd_stat`, and `ppd_stat_grouped`. The argument is by default set to `FALSE`, so the functions behave as previously. However, when it is set to `TRUE`, it calls `geom_bar` and plots a bar plot.

Tasks:
- [x] implement the argument
- [x] update tests
- [x] update documentation

## Examples

```r
y <- example_y_data
yrep <- example_yrep_draws
ppc_stat(y, yrep, stat = "median")

group <- example_group_data
ppc_stat_grouped(y, yrep, group, "median")
``` 
ppc_stat  ppc_stat_grouped
So, the cases presented in the documentation are working as they should.

Now, let's look at the example case of @santikka:
```r
y_discrete <- rbinom(20, 1, 0.2)
yrep_discrete <- matrix(rbinom(2000, 1, prob = 0.4), 1000, 20, byrow = TRUE)
ppc_stat(y_discrete, yrep_discrete, function(x) sum(abs(diff(x))))

group_discrete <- rep(c("First Half","Second Half"), each = 10)
ppc_stat_grouped(y_discrete, yrep_discrete, group_discrete, function(x) sum(abs(diff(x))))
``` 
ppc_stat_discrete_false  ppc_stat_group_discrete_false

On the other hand, when we use the `discrete` argument, they are plotted nicely as bar plots.
```r
ppc_stat(y_discrete, yrep_discrete, function(x) sum(abs(diff(x))), discrete = TRUE)
ppc_stat_grouped(y_discrete, yrep_discrete, group_discrete, function(x) sum(abs(diff(x))), discrete = TRUE)
``` 
ppc_stat_discrete_true  ppc_stat_grouped_discrete_true

A similar logic applies to `ppd_stat` versions as well.
```r
ppd_stat(yrep, "median")
ppd_stat_grouped(yrep, group, "median")
``` 
ppd_stat ppd_stat_grouped 

```r
ppd_stat(yrep_discrete, "median")
ppd_stat_grouped(yrep_discrete, group_discrete, "median")
``` 
ppd_stat_discrete_false  ppd_stat_grouped_discrete_false 

```r
ppd_stat(yrep_discrete, "median", discrete = TRUE)
ppd_stat_grouped(yrep_discrete, group_discrete, "median", discrete = TRUE)
``` 
ppd_stat_discrete_true ppd_stat_grouped_discrete_true


jgabry:
Great, thanks @behramulukir, this looks good.

PR #378, opened 2025-10-23
New pkgdown theme

VisruthSK:
Updating the theme to better match the main Stan page. This PR uses the new theme & use a GitHub Action to automatically build the dev version of the site on PRs and the release version on releases.

~~The action needs to be tested to see if all dependencies are properly specified.~~

### DIFF
--- a/R/helpers-ppc.R
+++ b/R/helpers-ppc.R
@@ -101,7 +101,7 @@ validate_predictions <- function(predictions, n_obs = NULL) {
 #' Validate PIT
 #'
 #' Checks that `pit` is numeric, doesn't have any NAs, and is either a vector,
-#' or a 1-D array with values in [0,1].
+#' or a 1-D array with values in `[0,1]`.
 #'
 #' @param pit The 'pit' object from the user.
 #' @return Either throws an error or returns a numeric vector.
@@ -521,8 +521,8 @@ get_interpolation_values <- function(N, K, L, prob) {
 #' within the bounds until z1 and takes the value in x1 at z1.
 #' @param x1 Vector of scaled ECDF values at the left end of the interval, z1.
 #' @param x2 Vector of scaled ECDF values at the right end of the interval, z2.
-#' @param z1 Left evaluation point in [0,1]
-#' @param z2 Right evaluation point in [0,1] with z2 > z1.
+#' @param z1 Left evaluation point in `[0,1]`
+#' @param z2 Right evaluation point in `[0,1]` with z2 > z1.
 #' @param N Total number of values in the sample.
 #' @return A vector containing the probability to transitioning from the values
 #' in x1 to each of the values in x2 weighted by the probabilities in p_int.
--- a/R/zzz.R
+++ b/R/zzz.R
@@ -10,3 +10,6 @@
                           "\nUse update.packages('ggplot2') to update.")
   }
 }
+.onLoad <- function(libname, pkgname) {
+  bayesplot_theme_set()
+}
--- a/tests/testthat/test-aesthetics.R
+++ b/tests/testthat/test-aesthetics.R
@@ -210,6 +210,18 @@ test_that("ggplot2::theme_set overrides bayesplot theme", {
   expect_identical(bayesplot_theme_get(), minimal)
 })
 
+test_that(".onLoad keeps ggplot theme and resets bayesplot theme", {
+  old <- ggplot2::theme_set(ggplot2::theme_minimal())
+  before <- ggplot2::theme_get()
+  bayesplot:::.onLoad()
+
+  expect_identical(bayesplot_theme_get(), default)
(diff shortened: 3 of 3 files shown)


jgabry:
Just started looking at this. First things I noticed looking at the GHA logs are that the new `yml` file needs to be in .Rbuildignore (there's an R cmd check NOTE about that) and that there are a lot of test failures. I'm thinking the test failures are unrelated since they're visual tests checking SVGs. I'll look into that. 

Update: test failures are definitely unrelated to this. I'm fixing that on a separate branch: https://github.com/stan-dev/bayesplot/pull/379

VisruthSK:
So tests seem to be ok, but I can't see the new pkgdown action, so we need to check that before merging.

jgabry:
> So tests seem to be ok, but I can't see the new pkgdown action, so we need to check that before merging.

Yeah tests should be fixed by the branch that I merged in. But I also don't see the pkgdown action.

VisruthSK:
It seems like the action must be run on PRs before it appears--otherwise it can only be seen on main

VisruthSK:
https://mc-stan.org/bayesplot/dev/

It's up now! Looks like some plots on the main page are broken so I'll work on that.

Plots on other pages, viz. a vignette I chcked, seem to be fine.

VisruthSK:
~~@jgabry images in the readme aer stored in the ./images subdir--is it ok to move them into ./man/figures instead? I'm guessing the reason for having them in a separate folder is so that they're not copied in when the package is installed~~

Fixed without having to move things. Should be ready for review.

jgabry:
Looking good! Just a couple of things I notice in the dev site: 

- The badge for R cmd check isn't displaying anymore. You can see this in the rendered dev site and also looking at the readme on the branch: https://github.com/stan-dev/bayesplot/blob/new-pkgdown-setup/README.md
- The ggplot theme used is different across the vignettes, which isn't the case in the current (release) website. 
   - https://mc-stan.org/bayesplot/dev/articles/plotting-mcmc-draws.html (uses default ggplot theme)  
   - https://mc-stan.org/bayesplot/dev/articles/visual-mcmc-diagnostics.html (uses the correct bayesplot default theme) 
   - https://mc-stan.org/bayesplot/dev/articles/graphical-ppcs.html (uses default ggplot theme) 

    It looks like in the one that uses the correct theme we explicitly set the theme via `bayesplot_theme_set()` and in the others we don't, so that could be the reason. But it shouldn't be necessary to explicitly set that (it happens internally) and I've never seen that happening when rendering the website before (it all looks fine at mc-stan.org/bayesplot). 

   I just pushed a commit explicitly setting the theme and if that works we can just not worry about this.

jgabry:
>   I just pushed a commit explicitly setting the theme and if that works we can just not worry about this.

Do we need to manually run the workflow to regenerate the dev site?

VisruthSK:
For now yes--disabled runs on PRs.

VisruthSK:
The badge and plots seem to be fixed.

jgabry:
For some reason the same ggplot2 theme issue is happening on the function documentation pages in the Examples sections. E.g., compare the plots at the bottom of these pages: 

* Dev: https://mc-stan.org/bayesplot/dev/reference/PPC-distributions.html#ref-examples
* Release: https://mc-stan.org/bayesplot/reference/PPC-distributions.html#examples

I guess we can manually set the theme in each examples section, but I don't understand why this needs to be done when it didn't need to be previously.  I think it must have to do with some change in ggplot2 v4.0 that's preventing bayesplot's default them from being set by default, that's the only thing I can think of.

jgabry:
> For some reason the same ggplot2 theme issue is happening on the function documentation pages in the Examples sections. E.g., compare the plots at the bottom of these pages:
> 
> * Dev: https://mc-stan.org/bayesplot/dev/reference/PPC-distributions.html#ref-examples
> * Release: https://mc-stan.org/bayesplot/reference/PPC-distributions.html#examples
> 
> I guess we can manually set the theme in each examples section, but I don't understand why this needs to be done when it didn't need to be previously. I think it must have to do with some change in ggplot2 v4.0 that's preventing bayesplot's default them from being set by default, that's the only thing I can think of.

I guess that can be fixed (if possible) separately and we can go ahead with the new site since it's (probably) not an issue with the site itself. Let me think about it for a min though.

VisruthSK:
~~Could this be a ggplot V4 thing?~~

Sorry reread your first comment. I'm curious if the same thing happens rendering the old site but bumping the ggplot dependency to 4.0.0

jgabry:
I think the same thing would happen. I'm 99% sure it's a ggplot v4.0 thing. I'm working on fixing it. I sort of have a fix, but there are some subtleties I'm still working out.

jgabry:
I just merged in a fix (I think) for the ggplot theme issue

PR #384, opened 2025-10-24
Restore pre ggplot2 v4.0 behavior of bayesplot theme setting

jgabry:
Ensures bayesplot keeps its own default theme by stripping .Environment metadata from ggplot themes before comparison, remembering ggplot’s baseline theme, and only switching to ggplot’s theme when it truly changes.


jgabry:
I think there are still some kinks to work out

PR #397, opened 2025-12-08
Testthat 3e

### LINKED ISSUE #381: Testthat V3
Swap to latest version of testthat. I haven't looked at the tests themselves, so I don't know if snapshots are appropriate anywhere--but 3e would allow for parallel testing which should be a big help. Also, bumping to the latest edition is probably a good idea as a matter of practice.

VisruthSK:
Changed syntax to match testthat third edition syntax (swapped to non-deprecated functions and so on.) Also gitignored .vscode formula for code-oss editors workspace settings.

Closes #381.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -63,5 +63,3 @@ VignetteBuilder: knitr
 Encoding: UTF-8
 Roxygen: list(markdown = TRUE)
 Config/testthat/edition: 3
-Config/testthat/parallel: true
-Config/testthat/start-first: mcmc-scatter-and-parcoord, ppc-distributions, mcmc-intervals, extractors


VisruthSK:
Doesn't seem like there is any speedup, but worth swapping to 3e anyway for updates. If runs consistently take longer than sequential, should probably swap back.

jgabry:
Great, thank you. I just triggered ubuntu-latest (devel) to run again since it failed for some reason. 

> If runs consistently take longer than sequential, should probably swap back.

Yeah might as well switch back if there's no benefit.

jgabry:
Hmm, still erroring only on ubuntu-latest (devel). Unclear why (not much to go on in the logs from what I can tell). Could maybe be a temporary GHA issue (those seem to happen from time to time). But I guess could also be something we need to fix.

jgabry:
I'm going to try disabling the parallel testing and see if that fixes it

VisruthSK:
Neat that seemed to have fixed it. Some system bug?

jgabry:
I guess, I'm not sure, it was pretty hard to tell anything useful from the logs. But glad it's fixed. The rest of this PR looks good. I'll go ahead and merge.

PR #399, opened 2025-12-08
Added cmdstanr to GHA

### LINKED ISSUE #398: Add cmdstanr to GHA workflows
Just noticed that tests were getting skipped in R CMD CHECK since cmdstanr isn't being installed on runners; I don't see any reason not to add it and run these tests in CI/CD.

```r
   ══ Skipped tests (1) ═══════════════════════════════════════════════════════════
  • {cmdstanr} is not installed (1): 'test-extractors.R:120:3'
```

VisruthSK:
Added cmdstanr as a dependency so some tests aren't skipped.

Used in both the R CMD CHECK and coverage actions.

Fixes #398.

### DIFF
--- a/.github/workflows/test-coverage.yaml
+++ b/.github/workflows/test-coverage.yaml
@@ -44,7 +44,7 @@ jobs:
           covr::to_cobertura(cov)
         shell: Rscript {0}
 
-      - uses: codecov/codecov-action@v5
+      - uses: codecov/codecov-action@v6
         with:
           # Fail if error if not on PR, or if on PR and token is given
           fail_ci_if_error: ${{ github.event_name != 'pull_request' || secrets.CODECOV_TOKEN }}
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # bayesplot (development version)
 
+* Replace `apply()` with `storage.mode()` for integer-to-numeric matrix conversion in `validate_predictions()`.
+* Fixed `is_chain_list()` to correctly reject empty lists instead of silently returning `TRUE`.
 * Added unit tests for `mcmc_areas_ridges_data()`, `mcmc_parcoord_data()`, and `mcmc_trace_data()`.
 * Added unit tests for `ppc_error_data()` and `ppc_loo_pit_data()` covering output structure, argument handling, and edge cases.
 * Added vignette sections demonstrating `*_data()` companion functions for building custom ggplot2 visualizations (#435)
--- a/R/helpers-mcmc.R
+++ b/R/helpers-mcmc.R
@@ -247,6 +247,9 @@ df_with_chain2array <- function(x) {
 #' @param x object to check
 #' @return TRUE or FALSE
 is_chain_list <- function(x) {
+  if (length(x) == 0) {
+    return(FALSE)
+  }
   check1 <- !is.data.frame(x) && is.list(x)
   dims <- try(sapply(x, function(chain) length(dim(chain))), silent=TRUE)
   if (inherits(dims, "try-error")) {
--- a/R/helpers-ppc.R
+++ b/R/helpers-ppc.R
@@ -73,12 +73,7 @@ validate_predictions <- function(predictions, n_obs = NULL) {
   }
 
   if (is.integer(predictions)) {
-    if (nrow(predictions) == 1) {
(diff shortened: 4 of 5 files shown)


VisruthSK:
The issue in the vignette (`visual-mcmc-diagnostics.Rmd`) seems to be a Mac specific error (RNG issue?), not sure why its only being caught on this branch.

jgabry:
> The issue in the vignette (`visual-mcmc-diagnostics.Rmd`) seems to be a Mac specific error (RNG issue?), not sure why its only being caught on this branch.

Yeah there's a part in that vignette where we want to show some diagnostic plots and we need enough divergences from HMC for the plot to make sense. So we have a check to make sure that happens. The problem is that on different systems the same seed can result in slightly different results (if hardware is different or toolchain is different, etc.). I haven't seen this happen in a while though. We might need to go in and try a different seed here: https://github.com/stan-dev/bayesplot/blob/94f2365f61b62d76c4ecdba5ac86042d506958ca/vignettes/visual-mcmc-diagnostics.Rmd#L157

jgabry:
> Added Stan r-universe and cmdstanr as a dependency so some tests aren't skipped.

Seems to still be skipping a test due to not finding cmdstanr: 

https://github.com/stan-dev/bayesplot/actions/runs/20016348876/job/57396448626?pr=399#step:7:195

VisruthSK:
Huh none of them are running, that's strange. I'll look into this

jgabry:
It's also fine to just not use cmdstanr on GHA if this takes a while to figure out. Not the highest priority (if you're looking for things to do I can find other ones if this ends up being a pain). But if you are going to try adding cmdstanr to Suggests like in your latest commit then I think you'll also have to add the `Additional_repositories` field in the DESCRIPTION file and point it to https://stan-dev.r-universe.dev/.

VisruthSK:
I think we're getting somewhere though--I think the last Windows run didn't skip any tests and code coverage did increase which I'm thinking (hoping?) is due to those extra tests running.

VisruthSK:
I was running into an issue with rstan relying on a deprecated boost RNG (l'ecuyer), so for testing I bumped the DESCRIPTION's rstan version to 2.36 and everything works (except for the mac divergence thing)

VisruthSK:
Seems like using `seed = 20251208` for `fit_cp` worked?

VisruthSK:
Tests don't seem to be skipped either, is it ok to bump the rstan dependency @jgabry? It should be possible to not update it and just figure out dependencies but it'll be a bit more of a pain. 2.36 is agressive too, just needs to be a version which resolved [this issue](https://github.com/stan-dev/stan/issues/3256).

jgabry:
The issue with bumping the RStan version in DESCRIPTION that CRAN only has 2.32.7. We can certainly bump the version to at least 2.32 but I'm not sure if that's sufficient (I guess we could force GHA to use a more recent version as long as we don't require that in DESCRIPTION). What I don't understand is why we're only having this issue with this PR. Other recent PRs passed all checks on GHA without issue. Any ideas?

VisruthSK:
Oh right I forgot 2.32.7 was the latest CRAN release. Hm. I added the Stan R universe repository so I'm guessing that's changing the generated lockfile and some dependencies are getting pinned to incompatible versions. Should undo the description change and run pak locally to see what versions it pins. I also need to read into what exactly `extra-repositories: 'https://stan-dev.r-universe.dev'` does to the repos option. We could also just build cmdstanr from source right?

Does the R universe host old versions or just the latest release and devel binaries? If its the latter that's probably the issue since I'd hazard the newest cmdstanr version would force some other packages to be pinned to newer versions too

jgabry:
> Does the R universe host old versions or just the latest release and devel binaries? If its the latter that's probably the issue since I'd hazard the newest cmdstanr version would force some other packages to be pinned to newer versions too

I'm not entirely sure, but I think just the latest.

VisruthSK:
https://github.com/stan-dev/rstan/issues/1152

If a new rstan CRAN version is going to be released that should fix the issues here, but I still not entirely sure why this is only happening now.

VisruthSK:
Looks like adding the multiverse to action resolved this.

VisruthSK:
No skipped test on MacOS, Windows, Ubuntu-devel--looks like everything is working as intended

PR #402, opened 2025-12-19
adding `mcmc_dots` and `mcmc_dots_by_chain` functions

behramulukir:
This PR adds `mcmc_dots` and `mcmc_dots_by_chain` functions which were requested by @avehtari. 

These functions are simply an adaptation of the previously implemented `ppc_dots` function to the MCMC space. As `ppc_dots` does, these functions require `ggdist` to be installed, so it checks whether that is installed, and if it is, then it utilises `ggdist::stat_dots` to create dot plots.

Tasks:
- [x] implement functions
- [x] implement tests
- [x] update documentation

## Examples

```r
x <- example_mcmc_draws()

mcmc_dots(x)
``` 
mcmc_dots(x)

```r
mcmc_dots(x, quantiles = 80)
``` 
mcmc_dots(x,quantiles=80)

```r
mcmc_dots_by_chain(x, regex_pars = "beta")
``` 
mcmc_dots_by_chain(x,regex_pars=beta)

### DIFF
--- a/man-roxygen/args-dots.R
+++ b/man-roxygen/args-dots.R
@@ -1,4 +1,5 @@
 #' @param quantiles For dot plots, an optional integer passed to
 #'   [ggdist::stat_dots()] specifying the number of quantiles to use for a
-#'   quantile dot plot. If `quantiles` is `NA` (the default) then all data
-#'   points are plotted.
+#'   quantile dot plot. If `quantiles` is `NA` then all data points are plotted.
+#'   The default is `quantiles=100` so that each dot represent one percent of
+#'   posterior mass.
--- a/man/MCMC-distributions.Rd
+++ b/man/MCMC-distributions.Rd
@@ -219,8 +219,9 @@ lines are drawn. Set to \code{NULL} to remove the lines.}
 
 \item{quantiles}{For dot plots, an optional integer passed to
 \code{\link[ggdist:stat_dots]{ggdist::stat_dots()}} specifying the number of quantiles to use for a
-quantile dot plot. If \code{quantiles} is \code{NA} (the default) then all data
-points are plotted.}
+quantile dot plot. If \code{quantiles} is \code{NA} then all data points are plotted.
+The default is \code{quantiles=100} so that each dot represent one percent of
+posterior mass.}
 }
 \value{
 A ggplot object that can be further customized using the \strong{ggplot2} package.
--- a/man/PPC-distributions.Rd
+++ b/man/PPC-distributions.Rd
@@ -202,8 +202,9 @@ function.)}
 
 \item{quantiles}{For dot plots, an optional integer passed to
 \code{\link[ggdist:stat_dots]{ggdist::stat_dots()}} specifying the number of quantiles to use for a
-quantile dot plot. If \code{quantiles} is \code{NA} (the default) then all data
-points are plotted.}
+quantile dot plot. If \code{quantiles} is \code{NA} then all data points are plotted.
+The default is \code{quantiles=100} so that each dot represent one percent of
+posterior mass.}
 
 \item{probs}{A numeric vector of probabilities controlling where quantile
 lines are drawn. Set to \code{NULL} to remove the lines.}
--- a/man/PPD-distributions.Rd
+++ b/man/PPD-distributions.Rd
@@ -113,8 +113,9 @@ function.)}
 
 \item{quantiles}{For dot plots, an optional integer passed to
 \code{\link[ggdist:stat_dots]{ggdist::stat_dots()}} specifying the number of quantiles to use for a
-quantile dot plot. If \code{quantiles} is \code{NA} (the default) then all data
-points are plotted.}
+quantile dot plot. If \code{quantiles} is \code{NA} then all data points are plotted.
+The default is \code{quantiles=100} so that each dot represent one percent of
+posterior mass.}
 
 \item{notch}{For the box plot, a logical scalar passed to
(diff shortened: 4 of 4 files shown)


jgabry:
Do you know why two of the new visual tests are failing? Maybe you just need to regenerate the snapshots? 

```
══ Failed tests ════════════════════════════════════════════════════════════════
── Failure ('test-mcmc-distributions.R:218:3'): mcmc_dots renders correctly ────
Snapshot of `testcase` has changed.

Backtrace:
    ▆
 1. └─vdiffr::expect_doppelganger("mcmc_dots (default)", p_base) at test-mcmc-distributions.R:218:3
 2.   ├─base::withCallingHandlers(...)
 3.   └─testthat::expect_snapshot_file(...)
── Failure ('test-mcmc-distributions.R:227:3'): mcmc_dots_by_chain renders correctly ──
Snapshot of `testcase` has changed.

Backtrace:
    ▆
 1. └─vdiffr::expect_doppelganger("mcmc_dots_by_chain (default)", p_base) at test-mcmc-distributions.R:227:3
 2.   ├─base::withCallingHandlers(...)
 3.   └─testthat::expect_snapshot_file(...)
```

jgabry:
I just regenerated the SVGs and pushed them. Let's see if it passes now.

behramulukir:
Sorry that I wasn't able to get back to this soon. In [this commit](https://github.com/stan-dev/bayesplot/commit/0fd6519af0e839898c3eede5fc4c24c05ae20262), I attempted to re-run `test-mcmc-distributions.R` and update the SVGs, but this resulted in all visual tests failing, so I had to revert to the previous commit.

Now with your update, everything looks okay to me, and tests pass as well, so I'll mark this PR as ready.

avehtari:
- I know other related plots have that y-axis, but could we drop it?
- I did not try to run, but looking at the code, it seems additional arguments will passed to `stats_dots`?

behramulukir:
- Is this what you mean by dropping the y-axis, @avehtari?

mcmc_dots_dropped_y

- Yes, additional arguments are passed to `stats_dots` in the same way as the `ppc_dots` function.

jgabry:
@avehtari Is there a reason why you want to drop the y-axis here but not for the other ones? Or do you eventually want to drop them for other plots too? I'm not necessarily opposed, but maybe we should do them all as part of one release? I guess this plot is new, so it doesn't matter so much...

avehtari:
Drop for all which have just one distribution and where y-axis has no purpose. Can be made in a separate PR

behramulukir:
Okay then, I'll update `mcmc_dots` so that it doesn't have a y-axis, and later we can make another PR for dropping it in other related functions. How does that sound to you, @jgabry?

jgabry:
Yeah that sounds good, we can get rid of the y-axis for `mcmc_dots()`. 

Dropping it for the existing plots is technically breaking backwards compatibility (visually at least). But I agree that it looks better without the y-axis so yeah we should consider doing that (in a separate PR)

jgabry:
> Dropping it for the existing plots is technically breaking backwards compatibility (visually at least). But I agree that it looks better without the y-axis so yeah we should consider doing that (in a separate PR)

I opened an issue for this.

avehtari:
Now that I actually checked out this branch and tested it, I would change this to use `quantiles=100` as default. 100 is great as then each dot represent 1% of posterior mass, and it's easy to examine tail quantiles. This is also what we recommend in https://teemusailynoja.github.io/visual-predictive-checks/

I really like that the vertical line is removed. This now means that it differs from `mcmc_hist()`, and I'm fine if the vertical line would be in this PR, and then we would remove it from all plots at once. But I guess it's fine that this doesn't have the vertical line as this is a new function and thus doesn't change the behavior for this function.

behramulukir:
Alright, I think I missed that `quantiles=100` is the recommended option in the paper. I updated `mcmc_dots` and `mcmc_dots_by_chain` to have it like that. Do you think we should also update default quantile values for `ppc_dots` and `ppd_dots`? This might affect people's existing plots. 

As we plan to drop the y-axis line from all similar plots in the future, I think it's okay if `mcmc_dots` doesn't have it in the release. Having a vertical line now and removing it in the future is unnecessary in my opinion.

jgabry:
> Alright, I think I missed that `quantiles=100` is the recommended option in the paper. I updated `mcmc_dots` and `mcmc_dots_by_chain` to have it like that. Do you think we should also update default quantile values for `ppc_dots` and `ppd_dots`? This might affect people's existing plots.

@avehtari what do you think about this question from @behramulukir? I guess that would technically break backwards compatibility, but the plots are still pretty new, so probably not a huge issue if we change those defaults. I think we should probably be consistent across all of the dot plots.

avehtari:
I prefer `quantiles=100` as the default in all these.

jgabry:
> I prefer `quantiles=100` as the default in all these.

@behramulukir Do you want to add that to this PR?

behramulukir:
@jgabry I updated the default quantiles for `ppc_dots()` and `ppd_dots()`. I think this PR should be ready to be merged once it passes the tests

jgabry:
Great, thanks @behramulukir!

jgabry:
I just updated the `args-dots` template so that it mentions the new default. I'll let the checks run again before merging but I think this is ready

behramulukir:
Yeah, thank you! I forgot to update that, sorry

jgabry:
No problem! Merging now

PR #412, opened 2026-01-04
Fix ggplot2 ‘Groups with fewer than two datapoints’ warning in mcmc_nuts_treedepth()

BjarkeHautop:
Currently, the following gives a warning (taken from test-mcmc-nuts.R)

```{r}
> if (requireNamespace("rstanarm", quietly = TRUE)) {
+     ITER <- 1000
+     CHAINS <- 3
+     fit <- rstanarm::stan_glm(mpg ~ wt + am, data = mtcars,
+                               iter = ITER, chains = CHAINS,
+                               refresh = 0)
+     np <- nuts_params(fit)
+     lp <- log_posterior(fit)
+ }
> 
> mcmc_nuts_treedepth(np, lp, chain = CHAINS)
Warning messages:
1: Groups with fewer than two datapoints have been dropped.
ℹ Set `drop = FALSE` to consider such groups for position adjustment purposes. 
2: Groups with fewer than two datapoints have been dropped.
ℹ Set `drop = FALSE` to consider such groups for position adjustment purposes.
```

This PR fixes the issue by filtering that there are at least 2 observations before calling `geom_violin`, so the dropping doesn't happen implicitly with a warning.

### DIFF
--- a/R/mcmc-diagnostics-nuts.R
+++ b/R/mcmc-diagnostics-nuts.R
@@ -369,19 +369,29 @@ mcmc_nuts_treedepth <- function(x, lp, chain = NULL, ...) {
     yaxis_ticks(FALSE)
 
   violin_lp_data <- data.frame(treedepth, lp = lp$Value)
+
+  # Only keep treedepth values that occur more than once for violin plot
+  value_counts <- table(violin_lp_data$Value)
+  keep_values <- names(value_counts[value_counts > 1])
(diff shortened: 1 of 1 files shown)


jgabry:
Great, thank you very much! Tests are running now.

jgabry (approved):
Changes look good and all tests are passing. Merging now. Thanks again!

PR #413, opened 2026-01-04
Clarify MCMC-nuts chain argument documentation for `mcmc_nuts_stepsize()`.

BjarkeHautop:
The default behavior for `mcmc_nuts_stepsize()` and what the argument `chain` does is different than for the rest. Updated the documentation to reflect this.


jgabry (approved):
Thanks for fixing this. I hadn't noticed this.

Issue #408, opened 2025-12-30
Deprecated ggplot2 arguments


BjarkeHautop:
Currently, there are 3 deprecated ggplot2 arguments that appears when running `devtools::test()`:

- The `draw_quantiles` argument of `geom_violin()` is deprecated as of ggplot2 4.0.0.
i Please use the `quantiles.linetype` argument instead. (Note this deprecation is supposed to say `quantile.linetype`, as seen in the documentation https://ggplot2.tidyverse.org/reference/geom_violin.html)

This is only used internally, since the public API for `ppc_violin_grouped()` uses `probs`. However, when changing all instances of `draw_quantiles` to `quantile.linetype`, a test in `test-ppc-distributions.R` at line 334/335 fails:

```
> ppc_violin_grouped(vdiff_y, vdiff_yrep, vdiff_group)
Error:
! Problem while converting geom to grob.
ℹ Error occurred in the 1st layer.
Caused by error in `draw_panel()`:
! `geom_path()` can't have varying colour, linewidth, and/or alpha along the line when linetype isn't solid.
Run `rlang::last_trace()` to see where the error occurred.
```

- `ppc_intervals()` / `ppd_intervals()` use deprecated ggplot2 arguments (`fatten`, `size`).

The `fatten` argument of `geom_pointrange()` is deprecated as of ggplot2 4.0.0.
i Please use the `size` aesthetic instead.

Using `size` aesthetic for lines was deprecated in ggplot2 3.4.0.
i Please use `linewidth` instead.

Changing the public API would be breaking. Furthermore, replacing fatten with size would mean the new size would semantically correspond to the old fatten, not the existing size.

jgabry:
Thanks @BjarkeHautop! @VisruthSK and I had noticed some of these recently with the release of ggplot2 v4.0. But you mention some additional details I hadn't considered (thank you). A few comments and questions: 

> This is only used internally, since the public API for ppc_violin_grouped() uses probs. However, when changing all instances of draw_quantiles to quantile.linetype, a test in test-ppc-distributions.R at line 334/335 fails:

[...]

PR #414, opened 2026-01-04
Fix all `size` warnings calls to ggplot2 (deprecated at 3.4.0). 

BjarkeHautop:
This PR fixes all the `size` warnings that ggplot2 threw (see also issue #408) by replacing them with `linewidth`.

I also added `size` as deprecated in `ppc_km_overlay()` and `p_overlay()` using `lifecycle::deprecated`.


jgabry:
Thank you! I definitely think we should change all internal cases of this. For the user-facing arguments, is there a reason only to do this for plots in `ppc-censoring.R`? For example, `mcmc_trace()` also has a user-facing `size` argument for controlling line width and I would guess that other functions do too (I haven't done a thorough check recently).

BjarkeHautop:
> For the user-facing arguments, is there a reason only to do this for plots in `ppc-censoring.R`? For example, `mcmc_trace()` also has a user-facing `size` argument for controlling line width and I would guess that other functions do too (I haven't done a thorough check recently).

No, we should probably do that. But note that some `size` arguments are still correct, so we can't just replace all `size` with `linewidth` without looking at the code.

jgabry:
Do mind removing the the deprecations in the ppc-censoring functions from this PR? Let's handle internal changes in this PR and then we can do the user-facing changes separately. I think that will make it easier to review and reason about. Is that ok with you? Thanks again for working on this!

BjarkeHautop:
No problem! Removed all deprecation stuff.

jgabry (approved):

jgabry:
Thank you! Looks good, merging now.

Issue #377, opened 2025-10-17
Feature request: ppc_rootogram_grouped()


mhollanders:
Hi all,

I'm enjoying the new rootogram PPC for count data. I was wondering if a `_grouped()` function was in the making?

Cheers,

Matt

jgabry:
Glad you like the new rootogram style and thanks for the suggestion. @behramulukir Any interest in working on this when you have time?

behramulukir:
Yeah, I would want to work on this. I don't know how soon I can do it, but I'll try to have a look at it

jgabry:
Thanks @behramulukir! Hope your semester is going well.

PR #419, opened 2026-01-28
adding `ppc_rootogram_grouped` function

behramulukir:
This PR adds `ppc_rootogram_grouped` function requested at #377. If you have any feedback, I am happy to hear @mhollanders.

`ppc_rootogram_grouped` simply creates multiple facets (plots) of the rootogram based on the grouping user passes to the function as an argument. 

Something to note here is that in `ppc_bars_grouped`, `force_axes_in_facets` function is used to make sure all facets have axes. This was resulting in errors when used with `ppc_rootogram_grouped(style="discrete")` since `-Inf` results in `NaN` when used with `scale_y_sqrt()`. It was still used in other rootogram styles, but so far, I hadn't managed to get it working for discrete style. I'll mark this PR as ready once I find a solution to that.

Tasks:
- [x] implement functions
- [x] implement tests
- [x] update documentation
- [x] fixing axes in `discrete` style
 
## Examples

```r
y <- rpois(100, 20)
yrep <- matrix(rpois(10000, 20), ncol = 100)
group <- gl(2, 50, length = 100, labels = c("GroupA", "GroupB"))

ppc_rootogram_grouped(y, yrep, group, prob = 0.5)
``` 

ppc_rootogram_grouped

```r
ppc_rootogram_grouped(y, yrep, group, prob = 0.5, style = "discrete")
``` 
ppc_rootogram_grouped_discrete

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,6 @@
 # bayesplot (development version)
 
+* Eliminate redundant data processing in `mcmc_areas_data()` by reusing the prepared MCMC array for both interval and density computation.
 * Validate equal chain lengths in `validate_df_with_chain()`, reject missing
   chain labels, and renumber data-frame chain labels internally when converting
   to arrays.
--- a/R/mcmc-intervals.R
+++ b/R/mcmc-intervals.R
@@ -599,64 +599,13 @@ mcmc_intervals_data <- function(x,
   prob <- probs[1]
   prob_outer <- probs[2]
 
-  x <- prepare_mcmc_array(x, pars, regex_pars, transformations)
-  x <- merge_chains(x)
-
-  data_long <- melt_mcmc(x) %>%
+  data_long <- melt_mcmc(
+    merge_chains(prepare_mcmc_array(x, pars, regex_pars, transformations))
+  ) %>%
     dplyr::as_tibble() %>%
     rlang::set_names(tolower)
 
-  probs <- c(0.5 - prob_outer / 2,
-             0.5 - prob / 2,
-             0.5 + prob / 2,
-             0.5 + prob_outer / 2)
-
-  point_est <- match.arg(point_est)
-  m_func <- if (point_est == "mean") mean else median
-
-  data <- data_long %>%
-    group_by(.data$parameter) %>%
-    summarise(
-      outer_width = prob_outer,
-      inner_width = prob,
-      point_est = point_est,
-      ll = unname(quantile(.data$value, probs[1])),
-      l  = unname(quantile(.data$value, probs[2])),
-      m  = m_func(.data$value),
-      h  = unname(quantile(.data$value, probs[3])),
-      hh = unname(quantile(.data$value, probs[4]))
-    )
-
-  if (point_est == "none") {
-    data$m <- NULL
-  }
-
-  color_by_rhat <- isTRUE(length(rhat) > 0)
-
-  if (color_by_rhat) {
-    rhat <- drop_NAs_and_warn(new_rhat(rhat))
-
-    if (length(rhat) != nrow(data)) {
-      abort(paste(
-        "'rhat' has length", length(rhat),
-        "but 'x' has", nrow(data), "parameters."
-      ))
-    }
-
-    rhat <- set_names(rhat, data$parameter)
-
-    rhat_tbl <- rhat %>%
-      mcmc_rhat_data() %>%
-      select(all_of("parameter"),
-             rhat_value = "value",
-             rhat_rating = "rating",
-             rhat_description = "description") %>%
-      mutate(parameter = factor(.data$parameter, levels(data$parameter)))
-
-    data <- dplyr::inner_join(data, rhat_tbl, by = "parameter")
-  }
-
-  data
+  compute_intervals(data_long, prob, prob_outer, point_est, rhat)
 }
 
 
@@ -691,16 +640,16 @@ mcmc_areas_data <- function(x,
   point_est <- match.arg(point_est)
   temp_point_est <- if (point_est == "none") "median" else point_est
 
-  intervals <- mcmc_intervals_data(x, pars, regex_pars, transformations,
-                                   prob = probs[1],  prob_outer = probs[2],
-                                   point_est = temp_point_est, rhat = rhat)
-
   x <- prepare_mcmc_array(x, pars, regex_pars, transformations)
   x <- merge_chains(x)
 
   data_long <- melt_mcmc(x) %>%
     dplyr::as_tibble() %>%
     rlang::set_names(tolower)
+
+  intervals <- compute_intervals(data_long, prob = probs[1],
+                                  prob_outer = probs[2],
+                                  point_est = temp_point_est, rhat = rhat)
 
   # Compute the density intervals
   data_inner <- data_long %>%
@@ -900,4 +849,62 @@ check_interval_widths <- function(prob, prob_outer) {
     warn(x)
   }
   sort(c(prob, prob_outer))
+}
+
+# Internal helper shared by mcmc_intervals_data() and mcmc_areas_data()
+compute_intervals <- function(data_long, prob, prob_outer,
+                              point_est = c("median", "mean", "none"),
+                              rhat = numeric()) {
+
+  probs <- c(0.5 - prob_outer / 2,
+             0.5 - prob / 2,
+             0.5 + prob / 2,
+             0.5 + prob_outer / 2)
+
+  point_est <- match.arg(point_est)
+  m_func <- if (point_est == "mean") mean else median
+
+  data <- data_long %>%
+    group_by(.data$parameter) %>%
+    summarise(
+      outer_width = prob_outer,
+      inner_width = prob,
+      point_est = point_est,
+      ll = unname(quantile(.data$value, probs[1])),
+      l  = unname(quantile(.data$value, probs[2])),
+      m  = m_func(.data$value),
+      h  = unname(quantile(.data$value, probs[3])),
+      hh = unname(quantile(.data$value, probs[4]))
+    )
+
+  if (point_est == "none") {
+    data$m <- NULL
+  }
+
+  color_by_rhat <- isTRUE(length(rhat) > 0)
+
+  if (color_by_rhat) {
+    rhat <- drop_NAs_and_warn(new_rhat(rhat))
+
+    if (length(rhat) != nrow(data)) {
+      abort(paste(
+        "'rhat' has length", length(rhat),
+        "but 'x' has", nrow(data), "parameters."
+      ))
+    }
+
+    rhat <- set_names(rhat, data$parameter)
+
+    rhat_tbl <- rhat %>%
+      mcmc_rhat_data() %>%
+      select(all_of("parameter"),
+             rhat_value = "value",
+             rhat_rating = "rating",
+             rhat_description = "description") %>%
+      mutate(parameter = factor(.data$parameter, levels(data$parameter)))
+
+    data <- dplyr::inner_join(data, rhat_tbl, by = "parameter")
+  }
+
+  data
 }


mhollanders:
Fantastic, thank you! I will doubtlessly have a use case for this in the next two weeks, and I will report if there's any issues.

mhollanders:
@behramulukir, sorry for the basic question, but how do I get to the function? I don't see it when I install the latest development version of bayesplot.

behramulukir:
Did you install the correct branch @mhollanders? As the changes are not merged to the master branch yet, you should install the branch I am working on. You can do it with `devtools::install_github("stan-dev/bayesplot", ref = "ppc-rootogram-grouped")`

mhollanders:
Thank you, and pardon for my ignorance. It works great!

behramulukir:
No worries! I'm glad that you liked it

jgabry:
Thanks @behramulukir and thanks @mhollanders for testing it out! 

> Something to note here is that in `ppc_bars_grouped`, `force_axes_in_facets` function is used to make sure all facets have axes. This was resulting in errors when used with `ppc_rootogram_grouped(style="discrete")` since `-Inf` results in `NaN` when used with `scale_y_sqrt()`. It was still used in other rootogram styles, but so far, I hadn't managed to get it working for discrete style. I'll mark this PR as ready once I find a solution to that.
> 

Does `0` work instead of `-Inf`? If `0` works here we could add an argument to `force_axes_in_facets()` to change between `-Inf` and `0` (leaving the default as `-Inf` so it stays the same for the other plots).

mhollanders:
It's tangentially related but I might as well piggy-back on to this post: are there any PPC tips for zero-inflated count data? This output is for occupancy models, where zero-inflation is modeled at the level of the site and species. So rare species will have lots of 0s. I'm attaching two plots, one for the aggregated (predicted) counts per species and site (*Q*), and one for the predicted counts per species, site, and replicate survey. (*y*) The excess zeros make them hard to interpret!

fig-ppc-Q
fig-ppc-y

behramulukir:
> Does `0` work instead of `-Inf`? If `0` works here we could add an argument to `force_axes_in_facets()` to change between `-Inf` and `0` (leaving the default as `-Inf` so it stays the same for the other plots).

The issue here is that `scale_y_sqrt` adds a padding before the x-axis line so that it is plotted at the negative y range. This exists in the original `ppc_rootogram(style="discrete")` as well. It is normally fine and looks visually good. The issue comes up when we try to plot the y-axis line in the negative y range. Since we cannot assign a negative y value, we can only plot the y-axis line up to y=0. This creates a gap between the end of the y-axis line and the x-axis line. Trying different values in `force_axes_in_facets()` yielded me results that work only in some cases and fail in other cases. For example, here is a half-working solution I was able to come up with:
```r
force_axes_in_facets_sqrt <- function() {
  thm <- bayesplot_theme_get()
  annotate(
    "segment",
    x = c(-Inf, -Inf), xend = c(0, -Inf),
    y = c(0, 0), yend = c(0, Inf),
    color = thm$axis.line$colour %||% thm$line$colour %||% "black",
    linewidth = thm$axis.line$linewidth %||% thm$line$linewidth %||% 0.5
  )
}
```  

As you can see in the following examples, there is a gap between the y-axis line and the x-axis line, and it fails in cases where we have multiple rows, since it doesn't plot an x-axis.

```r
ppc_rootogram_grouped(y, yrep, group, prob = 0.5, style = "discrete")
ppc_rootogram_grouped(y, yrep, group, prob = 0.5, style = "discrete", facet_args = list(nrow=2))
``` 

plot-76
plot-74


If we try to fix the issue of the lack of an x-axis line in multiple rows of plots, then we get two x-axis lines:

```r
force_axes_in_facets_sqrt <- function() {
  thm <- bayesplot_theme_get()
  annotate(
    "segment",
    x = c(-Inf, -Inf), xend = c(Inf, -Inf),
    y = c(0, 0), yend = c(0, Inf),
    color = thm$axis.line$colour %||% thm$line$colour %||% "black",
    linewidth = thm$axis.line$linewidth %||% thm$line$linewidth %||% 0.5
  )
}
``` 
plot-78
plot-77

This is simply because we cannot have any negative value at `y = c(0, 0), yend = c(0, Inf)`. The only viable solution I was able to find was removing the padding between the y=0 and the x-axis line by using `scale_y_sqrt(expand = expansion(mult = c(0,  0.05)))`.

plot-80
plot-81

If we decide to use this solution, we need either to change the look of the discrete rootogram slightly or to have a small difference between the grouped and normal versions of the discrete rootogram. I am not sure which one is better, though.

behramulukir:
> It's tangentially related but I might as well piggy-back on to this post: are there any PPC tips for zero-inflated count data?

Among the existing PPC count plots, I think `ppc_rootogram(style = "suspended")` is worth trying if you haven't tried it already. Since it plots the differences between expected and observed counts, it might be easier to interpret where the observed count is zero, and the expected count is non-zero (or vice versa).

You can also check [Visualizing Count Data Regressions Using Rootograms](https://www.zeileis.org/papers/Kleiber+Zeileis-2016.pdf) paper by Christian Kleiber and Achim Zeileis. The paper contains an example case where "a well-known data set from ethology, for which excess zeros and, more generally, overdispersion require treatment". 

If you find a solution that you think is worth implementing in bayesplot, please let me know! I would be interested in implementing it (unless you don't want to do it yourself, of course)

avehtari:
@mhollanders 

I would highly recommend using [Teemu's discrete rootogram](https://teemusailynoja.github.io/visual-predictive-checks/#fig-discrete-rootogram). Using square scaling on both x and y axis improves readability in case of high proportions and counts, and discrete nature of the data and model is more emphasized.

Also for zero-inflation part you should use binary target calibration plots as illustrated, for example, in [Nabiximols case study](https://users.aalto.fi/~ave/casestudies/Nabiximols/nabiximols.html#fig-calibration-0-betabinomial)

mhollanders:
Thanks! The rootograms in bayesplot are those rootograms, right?

behramulukir:
Yes, `ppc_rootogram(style = discrete)` is an adaptation of Teemu's discrete rootogram to the general style of bayesplot.

jgabry:
@behramulukir I had a chance to come back to this and I have two different ideas for handling the `force_axes_in_facets` issue. Let me know what you think. 

1. First idea is to define this helper function: 

```
force_axes_in_facets_sqrt <- function() {
  thm <- bayesplot_theme_get()
  annotate(
    "segment",
    x = c(-Inf, -Inf), xend = c(Inf, -Inf),
    y = c(0, 0), yend = c(0, Inf),
    color = thm$axis.line$colour %||% thm$line$colour %||% "black",
    linewidth = thm$axis.line$linewidth %||% thm$line$linewidth %||% 0.5
  )
}
```

Then inside the plotting function do something like this: 

```
  if (style == "discrete") {
    y_scale <- if (!is.null(dots$group)) {
      scale_y_sqrt(expand = expansion(mult = c(0, 0.05)))
    } else {
      scale_y_sqrt()
    }
...
```

and then use `y_scale` instead of calling `scale_y_sqrt()` directly. 

Then later: 

```
if (style == "discrete") {
  g <- g + force_axes_in_facets_sqrt()
} else {
  g <- g + force_axes_in_facets()
}
```

image

2. Another possibility: In ggplot2 3.5.0+ (we currently require 3.4.0+ but we could increase), `facet_wrap()` can draw interior axes natively via `axes` / `axis.labels`, so we can maybe avoid the annotation hack entirely for this case. Here is some rough code: 

```
bars_group_facets <- function(facet_args,
                              scales_default = "fixed",
                              force_axes = FALSE,
                              axis_labels_default = "all") {
  facet_args[["facets"]] <- "group"
  facet_args[["scales"]] <- facet_args[["scales"]] %||% scales_default
  
  if (force_axes) {
    facet_args[["axes"]] <- facet_args[["axes"]] %||% "all"
    facet_args[["axis.labels"]] <- facet_args[["axis.labels"]] %||%
      axis_labels_default
  }
  
  do.call("facet_wrap", facet_args)
}

# Then inside the plotting function: 
use_native_axes <- style == "discrete" && fixed_y(facet_args)
if (!use_native_axes) {
   g <- g + force_axes_in_facets()
}
g + bars_group_facets(
   facet_args,
   force_axes = use_native_axes,
   axis_labels_default = "margins"
)
```

One issue with this approach is we can't control the style of the interior axes, they just follow the general theme: 

image

behramulukir:
Thank you so much, @jgabry! I was also thinking about this and wasn't able to come up with anything new. It's great that you had fresh ideas.

I think your first idea is similar to what I proposed in the previous message, which didn't fully convince me.
> The only viable solution I was able to find was removing the padding between the y=0 and the x-axis line by using `scale_y_sqrt(expand = expansion(mult = c(0, 0.05)))`.

I personally prefer to have a gap between the dots and the x-axis line, and I felt like removing it led to worse results aesthetically and felt rather cluttered. It also broke the consistency of the graphs by having an unignorable visual difference between the grouped and non-grouped versions of the same plot.

However, I quite liked the second idea. I wasn't aware of the possibility of drawing interior axes natively, and I liked it. I think it looks nice in the image you shared. Compared to the interior axes of other grouped plots, it seems like the only difference is the existence of tick marks in the axis lines. I think that's a smaller difference than the gap between dots and the x-axis line, and the one which I would prefer.

What's your opinion, though? Do you have a preference between the two options?

jgabry:
You're right my first option is basically one of the options you already mentioned. If we like that option but don't like that the points at y=0 are cut off (that was the problem right? was there something else too?) then we could potentially nudge those points up a little bit by adding a small constant to those: 

image

What do you think about that? I think it looks good superficially, but is it maybe too confusing that the points aren't exactly at 0? Actually I think it should be possible to shift the y=0 tick mark and label up by the nudge amount to make it look like the points are exactly at 0! It would definitely be hack but it could work. 

For option 2, I think it works great and is not a hack. We just can't make the style of the added axes match the style of other similar plots that use force_axes_in_facets(). 

So I guess we have to compromise either way.

behramulukir:
> I think it looks good superficially, but is it maybe too confusing that the points aren't exactly at 0? Actually, I think it should be possible to shift the y=0 tick mark and label up by the nudge amount to make it look like the points are exactly at 0! It would definitely be hack, but it could work.

I believe it would look quite confusing without the little hack you suggested. However, since using that hack would solve the issue, I am positive about this workaround. That is especially if you feel like it wouldn't result in undesired consequences in edge cases.
 
> For option 2, I think it works great and is not a hack. We just can't make the style of the added axes match the style of other similar plots that use force_axes_in_facets().

To be honest, I am more positive about this solution. The main question I have is what changes do we apply to the style of the added axes compared to the general theme?
> One issue with this approach is we can't control the style of the interior axes, they just follow the general theme
```r
force_axes_in_facets <- function() {
  thm <- bayesplot_theme_get()
  annotate(
    "segment",
    x = c(-Inf, -Inf), xend = c(Inf,-Inf),
    y = c(-Inf,-Inf), yend = c(-Inf, Inf),
    color = thm$axis.line$colour %||% thm$line$colour %||% "black",
    linewidth = thm$axis.line$linewidth %||% thm$line$linewidth %||% 0.5
  )
}
``` 
Looking at the code above, it seems like the `force_axes_in_facets` is getting the colour and linewidth of the added axes from the general theme, right? If that's the case, the added axes following the style of the general theme shouldn't be an issue. Or am I missing something?

jgabry:
> Looking at the code above, it seems like the `force_axes_in_facets` is getting the colour and linewidth of the added axes from the general theme, right? If that's the case, the added axes following the style of the general theme shouldn't be an issue. Or am I missing something?

You're right, but when you look at the plots of the two options I put above the axes look different. The interior axis in option 1 is thinner than the main axis. I'm not entirely sure why, but this is what always happens with the interior axes when using the default bayesplot theme (maybe other themes too) with `force_axes_in_facets`. 

If we go with option 2 then the new interior axes have tick marks, which is inconsistent with all of the other plots that do use `force_axes_in_facets`. I don't think there's a way to remove those. But that's not too big of a deal. 

So I think if we accept that the interior axes for the facets will look a little different for this plot than the other plots then option 2 is preferable.

behramulukir:
> You're right, but when you look at the plots of the two options I put above the axes look different. The interior axis in option 1 is thinner than the main axis.

Yes, and that's why I was confused. Looking at the code, it seemed like there shouldn't be any difference, but it's visible that there is.

> So I think if we accept that the interior axes for the facets will look a little different for this plot than the other plots then option 2 is preferable.

Personally, I am fine with this as the difference is rather small, and it may be possible in future to tackle this inconsistency. However, if you feel like the consistency here is critical, I think that's a fair argument. In that case, I am also okay with going with the little hack you suggested 2 comments ago.

jgabry:
I’m ok with the small inconsistency too. I think it’s probably better not to use the hacky solution I came up with.

behramulukir:
Cool! I just increased the ggplot version to 3.5.0+ and added your solution for `ppc_rootogram_grouped`. It should pass tests and be ready to get merged

behramulukir:
Hey, no worries! I'm sorry that I rushed to make a commit. 

I think the solution for non-fixed y-scales is simple: if we use `force_axes_in_facets()` only when the style is not discrete, and make sure that ggplot draws interior axes in all cases when the style is discrete (through `bars_group_facets(force_axes = style == "discrete")`), we are supposed to get the correct plots. Here are some examples:

```r
ppc_rootogram_grouped(y, yrep, group, prob = 0.5, style = "discrete")
``` 
plot-1

```r
ppc_rootogram_grouped(
  y, 
  yrep, 
  group, 
  prob = 0.5, 
  style = "discrete", 
  facet_args = list(nrow = 2, scales = "free_y")
)
``` 
plot-2

```r
ppc_rootogram_grouped(
  y, 
  yrep, 
  group, 
  prob = 0.5, 
  style = "discrete", 
  facet_args = list(ncol = 2, scales = "free_y")
)
``` 
plot-3

```r
ppc_rootogram_grouped(
  y, 
  yrep, 
  group, 
  prob = 0.5, 
  style = "discrete", 
  facet_args = list(nrow = 2, scales = "free")
)
``` 
plot-4

```r
source("tests/testthat/data-for-ppc-tests.R")
ppc_rootogram_grouped(
  vdiff_y2,
  vdiff_yrep2,
  vdiff_group2,
  style = "discrete",
  facet_args = list(nrow = 2, scales = "free_y")
)
``` 

plot-5

I also added two tests for multi-row discrete cases, covering both free and fixed dimensions. Please let me know if there is anything still missing!

R/ppc-discrete.R
```diff
@@ -372,6 +388,36 @@ ppc_rootogram <- function(y,
...
+           size = 1,
+           bound_distinct = TRUE) {
+    check_ignored_arguments(...)
+    style <- match.arg(style)
+    call <- match.call(expand.dots = FALSE)
+    g <- eval(ungroup_call("ppc_rootogram", call), parent.frame())
+    
+    # In style = discrete, scale_y_sqrt() can't handle -Inf values in axis segments
+    if (style != "discrete") {
+      g <- g <- g + force_axes_in_facets()
```
jgabry:
```suggestion
      g <- g + force_axes_in_facets()
```

Issue #418, opened 2026-01-23
Remove y-axis for plots where y-axis has no purpose


jgabry:
This would technically break visual backwards compatibility although wouldn't affect anyone's code. @avehtari pointed out that for some plots we have a y-axis that serves no purpose and the plots look better without it. For example, the new `mcmc_dots()` function doesn't include the y-axis for this reason, but functions like `mcmc_dens()` and others include the y-axis for no reason.

LokeshShukla910:
I have pulled a commit request after solving the issue. Please review it and let me know if it need some changes

utkarshpawade:
Hi @jgabry, 

I've created PR #423 to address this issue. Since the previous PR appeared to be stalled, I've implemented the fix with the following changes:
- Removed y-axis line from all density, histogram, dot, violin, and frequency polygon plots (13 functions total across `mcmc_*`, `ppc_*`, and `ppd_*` families)
- Added `theme(axis.line.y = element_blank())` following the pattern already used in `mcmc_dots()`
- Updated all visual regression test snapshots

PR #420, opened 2026-02-07
Remove unnecessary y-axis from MCMC density and histogram plots

LokeshShukla910:
- Removed the y-axis text, title, and ticks from mcmc_dens(), mcmc_dens_overlay(), and mcmc_hist() plots, as the y-axis does not provide useful information for these visualizations.
- Updated the internal .mcmc_hist() and .mcmc_dens() functions to consistently suppress the y-axis for single and multi-parameter plots.
- This change improves plot aesthetics and consistency with mcmc_dots(), making the plots cleaner and easier to interpret.

### DIFF
--- a/R/mcmc-distributions.R
+++ b/R/mcmc-distributions.R
@@ -67,9 +67,11 @@
 #' }
 #' # example of using 'transformations' argument to plot log(sigma),
 #' # and parsing facet labels (e.g. to get greek letters for parameters)
-#' mcmc_hist(x, transformations = list(sigma = "log"),
-#'           facet_args = list(labeller = ggplot2::label_parsed)) +
-#'           facet_text(size = 15)
+#' mcmc_hist(x,
+#'   transformations = list(sigma = "log"),
+#'   facet_args = list(labeller = ggplot2::label_parsed)
+#' ) +
+#'   facet_text(size = 15)
 #' \donttest{
 #' # instead of list(sigma = "log"), you could specify the transformation as
 #' # list(sigma = log) or list(sigma = function(x) log(x)), but then the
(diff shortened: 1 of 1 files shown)


jgabry:
Hi, thank you for your interest in submitting a PR. Have you verified that the changes to the code work as intended? 

> Removed the y-axis text, title, and ticks from mcmc_dens(), mcmc_dens_overlay(), and mcmc_hist() plots, as the y-axis does not provide useful information for these visualizations.

These were already removed (see my review comments). What issue #418 is about is removing the axis line itself.  

> Updated the internal .mcmc_hist() and .mcmc_dens() functions to consistently suppress the y-axis for single and multi-parameter plots.

These changes don't suppress the y-axis. They turn off text and tick marks, but those were already disabled.

R/mcmc-distributions.R
```diff
@@ -555,6 +567,6 @@ mcmc_violin <- function(
     bayesplot_theme_get() +
     yaxis_text(FALSE) +
     yaxis_ticks(FALSE) +
-    yaxis_title(on = n_param == 1 && violin) +
```
jgabry, line 558: We need to keep this for the special case of 1 parameter _and_ a violin plot. If you look at a 1 parameter violin plot you'll see why (the y-axis provides the parameter name)

R/mcmc-distributions.R
```diff
@@ -237,7 +246,10 @@ mcmc_dens_overlay <- function(
     n_dens = n_dens,
     bounds = bounds,
     ...
-  )
+  ) +
+    yaxis_text(FALSE) +
+    yaxis_title(FALSE) +
+    yaxis_ticks(FALSE)
```
jgabry, line 252: These are already included in the internal `.mcmc_dens()` function so we don't need to add them again here. These don't turn off the y-axis, they just turn off the text and tick marks, but those were already off.

R/mcmc-distributions.R
```diff
@@ -170,7 +176,10 @@ mcmc_dens <- function(
     bounds = bounds,
     alpha = alpha,
     ...
-  )
+  ) +
+    yaxis_text(FALSE) +
+    yaxis_title(FALSE) +
+    yaxis_ticks(FALSE)
```
jgabry, line 182: These are already included in the internal `.mcmc_dens()` function so we don't need to add them again here. These don't turn off the y-axis, they just turn off the text and tick marks, but those were already off.

PR #424, opened 2026-02-24
Fix ppc-distribution docs for argument usage & correct test function (#366)

### LINKED ISSUE #366: fix ppc-distribution docs for argument usage & correct test function in test-ppc-distribution
In `ppc-distribution`, there is some insufficient information in the documentation, which should be updated. Also, `test-ppc-distribution` has a typo, which makes tests run the wrong function while testing. Here is a list of the things to be fixed:

documentation:
- `binwidth`, `bins`, and `breaks` arguments are passed to `ggdist::stat_dots` and `ggplot2::geom_area` in addition to `gggplot2::geom_hist`
- `freq` argument is used in frequency polygons in addition to histograms

testing:
- in `test-ppc-distribution`, `test_that("ppc_dens,pp_hist,ppc_freqpoly,ppc_boxplot return ggplot objects”)` function uses `ppc_dens` function instead of `ppd_dens` function at the part where it tests ppd versions of the functions

utkarshpawade:
Fixes #366

This PR addresses the documentation and testing issues in ppc-distribution:

### Documentation improvements:
- Clarified that `binwidth`, `bins`, and `breaks` arguments are passed to `ggplot2::geom_area()` and `ggdist::stat_dots()` in addition to `ggplot2::geom_histogram()`
- Clarified that `freq` argument applies to frequency polygons in addition to histograms

### Test fix:
- Fixed test in `test-ppc-distributions.R` that incorrectly used `ppc_dens()` instead of `ppd_dens()` when testing PPD versions of functions

### Testing:
- All tests in `test-ppc-distributions.R` pass (111 PASS | 0 FAIL)
- All documented functions work correctly with the updated parameters
- Documentation regenerated successfully

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # bayesplot (development version)
 
+* Improved documentation for `binwidth`, `bins`, and `breaks` arguments to clarify they are passed to `ggplot2::geom_area()` and `ggdist::stat_dots()` in addition to `ggplot2::geom_histogram()`
+* Improved documentation for `freq` argument to clarify it applies to frequency polygons in addition to histograms
+* Fixed test in `test-ppc-distributions.R` that incorrectly used `ppc_dens()` instead of `ppd_dens()` when testing PPD functions
 * New functions `mcmc_dots` and `mcmc_dots_by_chain` for dot plots of MCMC draws by @behramulukir (#402)
 * Default to `quantiles=100` for all dot plots by @behramulukir (#402)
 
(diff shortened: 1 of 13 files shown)


jgabry:
Thanks, this looks good. The unit tests are failing on ubuntu-latest (devel) but sometimes this happens on r-devel. We have visual tests that compare SVGs and they are very sensitive. Nothing here should affect those, so we can merge this.

jgabry (approved):

PR #428, opened 2026-03-04
Dependence aware tests for `ppc*ecdf`

florence-bockting:
## Description
The current approach in `ppc_loo_pit_ecdf` and `ppc_pit_ecdf` assumes independence of LOO-PIT values which is not valid (Marhunenda et al., 2005). The corresponding graphical test yields an envelope that is too wide, reducing the test's ability to reveal model miscalibration.
Tesso & Vehtari (2026, see [preprint](https://arxiv.org/abs/2603.02928)) propose three testing procedures that can handle any dependent uniform values and provide an updated graphical representation that uses color coding to indicate influential regions or most influential points of the ECDF. This PR implements the new development, by adding the updated approach (method = "correlated") additionally to the previous approach (method = "independent").

## TODOs
+ [x] updated `ppc_loo_pit_ecdf()` function in `ppc-loo.R`
+ [x] updated `ppc_pit_ecdf()` and `ppc_pit_ecdf_grouped()` function in `ppc-distributions.R`
+ [x] add unittests and visual regression tests
+ [x] update documentation
+ [x] deprecation suggestion of old method 
    + P1: old method (default) and new method available via `method` argument
    + P2: new method (default) and old method available via `method` argument
    + P3: old method is removed


jgabry:
I'm seeing the failure on R-devel outside of this PR too (there seem to be some tiny insignificant differences to the SVGs on r-devel, which has happened before). I haven't seen the Mac failure on r-release before, but it's possible it's not due to this PR. We'd need to check that one more closely.

jgabry:
Because this copies a lot of code from https://github.com/stan-dev/posterior/pull/435 and other parts of the posterior package, I think we should wait to merge it until that posterior PR is merged so we can remove the extra code before adding it to bayesplot. 

Related to this, I spoke with @avehtari and we think that it could be useful to have all the uniformity test code in posterior also because it could be used without making a plot. So that code could be added to https://github.com/stan-dev/posterior/pull/435 and then used in bayesplot.

florence-bockting:
> Because this copies a lot of code from [stan-dev/posterior#435](https://github.com/stan-dev/posterior/pull/435) and other parts of the posterior package, I think we should wait to merge it until that posterior PR is merged so we can remove the extra code before adding it to bayesplot.

Yes, I am totally with you.

> Related to this, I spoke with @avehtari and we think that it could be useful to have all the uniformity test code in posterior also because it could be used without making a plot. So that code could be added to [stan-dev/posterior#435](https://github.com/stan-dev/posterior/pull/435) and then used in bayesplot.

Alright. I moved the respective code from this PR to [stan-dev/posterior#435](https://github.com/stan-dev/posterior/pull/435)

florence-bockting:
As mentioned in the PR introduction text, the idea would be migration in multiple stages.
First stage is non-breaking behavior: That is users can simply use "original" code but are informed that this behavior will change in future.  Future changes consider changing the default method and thus are breaking.

Informing the user in the first stage is currently implemented as follows:

```
# using the "original" code provides a message about the change
> ppc_loo_pit_ecdf(y, yrep, lw)
ℹ In the next major release, the default `method` will change to 'correlated'.
• To silence this message, explicitly set `method = 'independent'` or `method = 'correlated'`.

# explicit use of "independent" method provides information that this method is superseded
> ppc_loo_pit_ecdf(y, yrep, lw, method = "independent")
The 'independent' method is superseded by the 'correlated' method.
```

R/ppc-loo.R
```diff
@@ -419,63 +446,252 @@ ppc_loo_pit_ecdf <- function(y,
...
+    }
+  }
+
+  switch(method,
+    "correlated" = {
+      if (!is.null(interpolate_adj)) .warn_ignored("'correlated'", "interpolate_adj")
+      
+      test <- match.arg(test %||% "POT", choices = c("POT", "PRIT", "PIET"))
+      alpha <- 1 - prob
+      gamma <- gamma %||% 0
```
jgabry: The doc says "If `NULL`, automatically determined based on p-value" but this just sets it to 0 if `NULL`. Is that intentional?

jgabry: Doesn't this mean that right now by default we highlight every positive contribution point/segment whenever `p < alpha`? Maybe that's what we want, but that's more than what's documented, right? Or am I misunderstanding? (quite possibly!)

R/ppc-loo.R
```diff
@@ -419,63 +446,252 @@ ppc_loo_pit_ecdf <- function(y,
...
+      linewidth <- linewidth %||% 0.3
+      color <- color %||% c(ecdf = "grey60", highlight = "red")
+      help_text <- help_text %||% TRUE
+      pareto_pit <- pareto_pit %||% is.null(pit) && test %in% c("POT", "PIET")
```
jgabry: I think R will parse this as 

```r
(pareto_pit %||% is.null(pit)) && (test %in% c("POT", "PIET"))
```

and _not_ as 

```r
pareto_pit %||% (is.null(pit) && test %in% c("POT", "PIET"))
```

So if we have 

```r
pareto_pit = TRUE
pit = NULL
test = "PRIT"
```

then I think this will parse as `TRUE && FALSE` -> `FALSE`. 

Do we not want this instead? 

```r
pareto_pit <- pareto_pit %||% (is.null(pit) && test %in% c("POT", "PIET"))
```

This is a bit confusing for me, so I could definitely be wrong!

R/ppc-loo.R
```diff
@@ -419,63 +446,252 @@ ppc_loo_pit_ecdf <- function(y,
...
-    N = n_obs,
-    K = K,
-    prob = prob,
-    interpolate_adj = interpolate_adj
+  unit_interval <- seq(0, 1, length.out = K)
+  ecdf_pit_fn <- ecdf(pit)
+  y_label <- if (plot_diff) "ECDF difference" else "ECDF"
+
+  if (method == "correlated") {
+    test_res <- posterior::uniformity_test(pit = pit, test = test)
```
jgabry: This assumes that a pvalue and pointwise are always returned, but can't `uniformity_test` error in some cases? E.g. 

```r
posterior::uniformity_test(c(0.2, 0.8), "POT")
posterior::uniformity_test(c(0.25, 0.5, 0.75), "POT")
posterior::uniformity_test(0.5, "POT")
```  

How should we handle this? Or are we unlikely to encounter it in the wild, so to speak?

jgabry (commented):
Here are a few comments and questions. So far I've only looked at some of the code, so I may have more, but I thought I would give you these now so we could start discussing.

R/ppc-loo.R
florence-bockting: > The doc says "If `NULL`, automatically determined based on p-value" but this just sets it to 0 if `NULL`. Is that intentional?

Thanks, this was wrong in the documentation. I updated it: "If `NULL` (default), `gamma` is set to 0, and thus all suspicious points are flagged."

R/ppc-loo.R
florence-bockting: Thank you for catching this behavior! I fixed the notation and created corresponding unittests. 
Furthermore, I noticed that I need to catch the situation where a user provides both `pareto_pit = TRUE` and `pit = pit` (thus, pit is not NULL). I added an error with the user message:

```
`pareto_pit = TRUE` cannot be used together with a non-`NULL` `pit` value. Set either `pareto_pit = FALSE` or `pit = NULL`.
```

R/ppc-loo.R
florence-bockting: Regarding `uniformity_test` : "POT" and "PRIT" are based on a truncated Cauchy combination test. That is, they error when no p-values below 0.5 exist; I am not aware of any condition where they would return NA. "PIET" is based on an untruncated version of CCT, it can return pvalue=NA under special conditions (which yields in an error later on in `ppc_*_ecdf`.

In the following, I worked out some edge cases (I don't know how likely this situation would actually happen "in the wild"):

```
# fails when test = "PIET" as pvalue=NA 
ppc_loo_pit_ecdf(pit = c(0, 1, 0.5), method = "correlated", test = "PIET")
# works when test="POT" or "PRIT" OR method="independent"
ppc_loo_pit_ecdf(pit = c(0, 1, 0.5), method = "correlated", test = "POT")
ppc_loo_pit_ecdf(pit = c(0, 1, 0.5), method = "independent")

# fails when test "POT" or "PRIT" as no pvalues < 0.5 exist
ppc_loo_pit_ecdf(pit = c(0.8, 0.5, 0.3, 0.1), method = "correlated", test = "POT")
# works when test="PIET" OR method="independent"
ppc_loo_pit_ecdf(pit = c(0.8, 0.5, 0.3, 0.1), method = "correlated", test = "PIET")
ppc_loo_pit_ecdf(pit = c(0.8, 0.5, 0.3, 0.1), method = "independent")
```

Now, I am uncertain what a clean suggestion for a user from a methodological point of view is (@avehtari): 

- changing the test type (e.g., from POT to PIET or the other way around)
- allowing to change truncation behavior for the test in case of failing
- something else

R/ppc-loo.R
florence-bockting: > Doesn't this mean that right now by default we highlight every positive contribution point

Yes, this is intended as it is the safer and more informative choice.

jgabry:
@florence-bockting The changes look good, but I think there are the same issues (with the `gamma` doc and the `pareto_pit` issue) in the ppc-distributions.R file too, right? Also, how much of the doc in the two R files now overlaps? Could we write some of it once and share it? (If it's only a small amount of overlap then maybe not worth it).

tests/testthat/_snaps/ppc-distributions/ppc-pit-ecdf-color-change.svg
avehtari, line 1: The p-value line should have a fint size similar to the axis tick label font. Now it's much smaller and hard to read

florence-bockting:
@jgabry
Thanks for the [comments](https://github.com/stan-dev/bayesplot/pull/428#issuecomment-4238680376). I was planning to do the refactoring for reducing duplicate code between the ppc-ecdf functions at a later stage as this was already in the original version the case. But given your comment I realized that I should probably do it just now. Thus:

I have refactored the code such that it removes duplicate code and documentation by creating new helper functions or template files.
Furthermore, I increased the helper-text size but also added now the factor for controlling size as an additional argument. When we improve the theme, we can probably get rid of it again. @avehtari

R/ppc-loo.R
florence-bockting: @jgabry Alright, I checked this point with Aki and added corresponding checks for the PIT values where a problem can occur in practice. Specifically, we realized that the failing examples rather pointed to a check whether 0,1 are included in the PIT values, which they should not when we expect continuous PIT values (for POT and PIET)

jgabry:
@florence-bockting please also officially add yourself to the DESCRIPTION file as part of this PR!

florence-bockting:
@jgabry just as a little reminder that this PR is still waiting for final review. Thank you

R/helpers-ppc.R
```diff
@@ -583,6 +583,352 @@ ecdf_intervals <- function(gamma, N, K, L = 1) {
...
+    bayesplot_theme_get()
+}
+
+#' @noRd
+.compute_pit_values <- function(y, yrep, lw, psis_object, group, K, pareto_pit,
+  pit, loo_cv
+) {
+  # pareto-pit values
+  if (isTRUE(pareto_pit) && is.null(pit)) {
+    suggested_package("rstantools")
```
jgabry: I don't think rstantools is actually needed for this branch here, right?

R/helpers-ppc.R
```diff
@@ -583,6 +583,352 @@ ecdf_intervals <- function(gamma, N, K, L = 1) {
...
+    ))
+    method <- "independent"
+  } else {
+    method <- rlang::arg_match(method, values = c("independent", "correlated"))
+    if (identical(method, "independent")) {
+      inform("The 'independent' method is superseded by the 'correlated' method.")
+    }
+  }
+
+  alpha <- 1 - prob
```
jgabry, line 630: Should we validate that `prob` is a valid probability? Either here or before it's passed in? I guess whatever is consistent with how we've been checking other arguments.

man-roxygen/args-pit-ecdf.R
```diff
@@ -1,19 +1,24 @@
...
+#' @param test When `method = "correlated"`, which dependence-aware test to use:
+#'   `"POT"`, `"PRIT"`, or `"PIET"`. Defaults to `"POT"`.
+#' @param gamma When `method = "correlated"`, tolerance threshold controlling
+#'   how strongly suspicious points are flagged. Larger values (`gamma > 0`)
+#'   emphasize points with larger deviations. If `NULL`, defaults to `0` and
+#'   thus all suspicious points are flagged.
+#' @param color When `method = "correlated"`, a vector with base color and
+#'   highlight color for the ECDF plot. Defaults to
+#'   `c(ecdf = "grey60", highlight = "red")`. The first element is used for
+#'   the main ECDF line, the second for highlighted suspicious regions.
```
jgabry: The code that processes the color vector requires these exact names to be used in order to work properly (I think) but it doesn't check for the names. It indexes `color["ecdf"]` and `color["highlight"]`. The doc also mentions "first element" and "second element", but it's the names, not the position, that are used in the code. The user can pass incorrect names or unnamed vectors and I think the colors won't render properly. We should probably either check for the names or just use the position (first element and second element) and forget about the names.

R/ppc-loo.R
```diff
@@ -400,23 +403,31 @@ ppc_loo_pit_qq <- function(y,
...
+#'   when `method = 'independent'`. The default is `FALSE`, but for large
+#'   samples we recommend setting `plot_diff = TRUE` to better use the plot area.
+#' @param interpolate_adj A boolean defining if the simultaneous confidence
+#'   bands should be interpolated based on precomputed values rather than
+#'   computed exactly. Computing the bands may be computationally intensive and
+#'   the approximation gives a fast method for assessing the ECDF trajectory.
+#'   For `ppc_loo_pit_ecdf()` when `method = 'independent'`.
+#'   The default is to use interpolation if `K` is greater than 200.
+#' @param pit An optional vector of probability integral transformed values for
+#'   which the ECDF is to be drawn. For `ppc_loo_pit_ecdf()`. If `NULL`, PIT
```
jgabry: Should we also mention the other `ppc_loo_*` functions that take the `pit` argument in addition to `ppc_loo_pit_ecdf()`?

R/ppc-distributions.R
```diff
@@ -670,78 +671,161 @@ ppc_violin_grouped <-
...
+#' @param pit An optional vector of probability integral transformed values for
+#'   which the ECDF is to be drawn. For `ppc_pit_ecdf()` and
+#'   `ppc_pit_ecdf_grouped()`. If `NULL`, PIT values are computed to `y` with
+#'   respect to the corresponding values in `yrep`.
+#' @param linewidth When `method = "correlated"`, the line width of the ECDF.
+#'   Defaults to `0.3`.
+#' @note
+#' Note that the default "independent" method is **superseded** by
+#' the "correlated" method (Tesso & Vehtari, 2026) which accounts for dependent
+#' PIT values.
```
jgabry: Because this is documented on the same page as the other function that the same `@note` it shows up twice in the .Rd file. However, it also doesn't show up at all (at least for me) when rendering the doc in RStudio. Maybe the Note field isn't rendered? Do you see that too? 

Another option would be to remove both uses of `@note` and instead use the Plot Descriptions section. And I guess same thing for the loo versions, if applicable.

jgabry (commented):
Here are some review comments. Sorry for the delay! Overall I think this is in good shape, just a few relatively minor comments.

R/helpers-ppc.R
florence-bockting: True. I removed it from this branch.

R/helpers-ppc.R
florence-bockting: I included a validation check in `.pit_ecdf_resolve_method_args` (`helpers-ppc.R`)  and included additional tests for validating the input check.

man-roxygen/args-pit-ecdf.R
florence-bockting: Good point. I would stay with the named vector as I gives the construct some structure than just providing two values in a vector. Therefore I

- updated the argument description, 
- included a check for the color input format and 
- added additional tests to validate the input check.

R/ppc-loo.R
florence-bockting: Yes, makes sense I included a reference to the other ppc_loo_* functions.

R/ppc-distributions.R
florence-bockting: True. In my case it actually renders the note information two times. 
I decided to remove the note section and include the "note information" into the corresponding Plot Description section. This makes anyway more sense as it applies only to the specific pit_ecdf functions and not to all functions listed on the page.

florence-bockting:
> Here are some review comments. Sorry for the delay! Overall I think this is in good shape, just a few relatively minor comments.

Thank you very much for your review. As always super helpful! I made changes corresponding to all your comments.

jgabry (approved):
I think this looks good! @avehtari any other comments/suggestions, or ready to merge?

PR #430, opened 2026-03-09
Deprecate user-facing size and fatten arguments for ggplot2 4.0 compatibility

### LINKED ISSUE #408: Deprecated ggplot2 arguments

ishaan-arora-1:
Fixes #408

Picks up the remaining tasks from #408 as outlined by @jgabry in https://github.com/stan-dev/bayesplot/issues/408#issuecomment-3709850055:

1. **Deprecate user-facing `size` arguments that control line width** — added a `linewidth` parameter across ~25 plotting functions (density overlays, ribbons, ECDFs, boxplots, violins, error binned, parcoord, acf, KM overlays). The old `size` argument still works but emits a deprecation warning via `lifecycle::deprecate_warn()`.

2. **Deprecate `fatten` for `geom_pointrange`** — `ppc_intervals()`, `ppd_intervals()`, `ppc_loo_intervals()`, `ppc_bars()`, and their grouped variants now use `fatten = deprecated()`. Point size is controlled directly by `size`, matching ggplot2 4.0 semantics. The default `size` was changed from 1 to 2.5 to preserve the previous visual appearance (`old size * fatten = 1 * 2.5`).

Added `lifecycle` as an imported dependency and created internal helpers (`resolve_linewidth()`, `resolve_fatten()`) to keep the deprecation logic consistent. Updated all roxygen docs and examples to use the new parameter names.


ishaan-arora-1:
@jgabry any comments on this pr?

jgabry:
I haven't had a chance to go through it yet. This one changes a lot of files and changes user-facing behavior, so I need to find some time to really look at it and test it. I've been trying to look at smaller PRs first and get those finished before I look at the bigger ones. I'm happy to have this PR! just need to find some time for it.

R/helpers-shared.R
```diff
@@ -45,6 +45,52 @@ check_ignored_arguments <- function(..., ok_args = character()) {
...
+  if (!is.null(size)) {
+    lifecycle::deprecate_warn(
+      when = "1.16.0",
+      what = paste0(calling_fn %||% "fn", "(size)"),
+      with = paste0(calling_fn %||% "fn", "(linewidth)")
+    )
+    return(size)
+  }
+  linewidth %||% default_linewidth
+}
```
jgabry, line 73: If a user supplies both `size` and `linewidth`, this helper prefers the deprecated `size` argument and ignores the preferred `linewidth`. I think we should error when both are supplied with an informative error message telling the user to only use `linewidth`.

R/ppc-discrete.R
```diff
@@ -277,6 +279,7 @@ ppc_rootogram <- function(y,
                           ...,
                           prob = 0.9,
                           size = 1,
+                          linewidth = 1,
```
jgabry: This (and how it's used below) changes `ppc_rootogram()` semantics without any deprecation warning. Before this PR, `size` controlled line width here but after this change, `ppc_rootogram(size = 3, style = "hanging")` still renders with linewidth `1`, and there is no warning telling the user to switch to `linewidth`.

NEWS.md
```diff
@@ -1,5 +1,7 @@
 # bayesplot (development version)
 
+* Deprecate user-facing `size` arguments that controlled line width in favor of linewidth` across all plotting functions.
```
jgabry:
```suggestion
* Deprecate user-facing `size` arguments that controlled line width in favor of `linewidth` across all plotting functions.
```

NEWS.md
```diff
@@ -1,5 +1,7 @@
...
+* Deprecate the `fatten` argument in `ppc_intervals()`, `ppd_intervals()`,`ppc_loo_intervals()`, `ppc_bars()`, and their grouped variants. Point size in`geom_pointrange()` is now controlled directly by `size`.
```
jgabry:
```suggestion
* Deprecate the `fatten` argument in `ppc_intervals()`, `ppd_intervals()`, `ppc_loo_intervals()`, `ppc_bars()`, and their grouped variants. Point size in `geom_pointrange()` is now controlled directly by `size`.
```

R/bayesplot-package.R
```diff
@@ -4,6 +4,7 @@
 #' @aliases bayesplot
 #'
 #' @import ggplot2 stats rlang
+#' @importFrom lifecycle deprecated deprecate_warn is_present
```
jgabry, line 7: This line is already present on line 9 and can be removed

jgabry (changes requested):
Thanks again for this PR! Here are some initial review comments (I may have more after spending some more time with this). Beyond these I think we need to think a bit more about what the new default for `size` should be when it replaces `fatten`. All of the plots using pointrange have much bigger points now (see the SVGs for ppc_bars, ppc_intervals, ppc_loo_intervals, ppd_intervals, and others) so we need to make those at least close to what they were before (doesn't have to be exact, but we can't have such a drastic change). When we were discussing this in the issue we thought it made sense to use size = 2.5 but that seems to have been an incorrect assumption. Maybe it's worth seeing what the plots look like with size = 1?

ishaan-arora-1:
@jgabry the issue is that it still multiplies size * fatten internally in geom pt range, and when fatten isn't supplied, it defaults to 4. So size=2.5 was producing points at visual size 2.5 * 4 = 10, way bigger than the old size=1, I've Changed the default back to size = 1 across all pointrange functions, which gives visual point size 1 * 4 = 4 which is slightly bigger than before but reasonable.

jgabry:
> @jgabry the issue is that it still multiplies size * fatten internally in geom pt range, and when fatten isn't supplied, it defaults to 4. So size=2.5 was producing points at visual size 2.5 * 4 = 10, way bigger than the old size=1, I've Changed the default back to size = 1 across all pointrange functions, which gives visual point size 1 * 4 = 4 which is slightly bigger than before but reasonable.

I see, that makes sense, thanks for the explanation. This does make me wonder if we should be using `geom_linerange` and `geom_point` together instead of `geom_pointrange`. I'm not sure. That would give us better control, right? What do you think? 

Also, I think we should take `ppc_rootogram` out of this PR (sorry I had you make changes to it just to then ask you to remove it). The rootogram is different because size currently has mixed semantics: it affects line width in hanging/suspended styles and also point appearance in discrete style. I think that will need it's own design and we should figure out how to handle that separately from this PR.

PR #432, opened 2026-03-10
Fix mcmc_areas_ridges() overlay breaking with + scale_y_discrete()

### LINKED ISSUE #287: mcmc_areas_ridges don't overlay when using scale_y_discrete
Hi, 

First - thanks to all developers for creating such a useful plotting package. 

I have come across a bug in mcmc_areas_ridges recently, which means that the density plots created in mcmc_areas_ridges don't overlay properly when used in conjunction with scale_y_discrete. 
I think this worked in the past, since I was reusing code from an earlier project which didn't have this problem - although I can't be sure. 

Here is what the plot looks like without using scale_y_discrete
image

And here is the same plot, using scale_y_discrete to reorder the categories in the y axis
image

here is a working example:
```
library(bayesplot)

n = 100

posterior = 
  data.frame(x = rnorm(n),
             y = rnorm(n),
             z = rnorm(n),
             a = rnorm(n))

# Overlay works properly
mcmc_areas_ridges(posterior, prob = 0.89,
                  pars = c("x", "y", "z"))

# Use scale_y_discrete to reorder y axis means overlapping fails
mcmc_areas_ridges(posterior, prob = 0.89,
                  pars = c("x", "y", "z")) + 
  scale_y_discrete(labels = c("z", "y", "x"))

# Don't reorder, but still use scale_y_discrete - overlapping fails
mcmc_areas_ridges(posterior, prob = 0.89,
                  pars = c("x", "y", "z")) + 
  scale_y_discrete(labels = c("x", "y", "z"))

# final test that should do nothing
mcmc_areas_ridges(posterior, prob = 0.89) + 
  scale_y_discrete(labels = c("z", "y", "x", "a"))
```

utkarshpawade:
Fixes #287

`mcmc_areas_ridges()` used `scale_y_discrete(limits = ...)` internally to set the y-axis draw order. When users added their own `+ scale_y_discrete(...)` (eg to rename labels), ggplot2 replaced the entire scale (including the limits) which flipped the draw order and broke ridgeline masking.

This fix encodes the draw order directly in the factor levels of
`data$parameter` (reversed so the first parameter appears on top), instead of relying on `limits =`. User-added `scale_y_discrete()` calls can now change labels without affecting the overlay order.

Changes:
- `mcmc-intervals.R`: reverse factor levels before building layers,
  remove `limits =` from `scale_y_discrete()`, update `par_draw_order`
- `test-mcmc-intervals.R`: add unit tests and `vdiffr` snapshot for the
  previously failing case

### DIFF
--- a/tests/testthat/test-mcmc-intervals.R
+++ b/tests/testthat/test-mcmc-intervals.R
@@ -89,14 +89,13 @@ test_that("mcmc_areas_ridges overlay is not broken by scale_y_discrete (#287)", 
   expect_gg(p_base + ggplot2::scale_y_discrete(labels = c("A", "B", "C")))
   expect_gg(p_base + ggplot2::scale_y_discrete(labels = pars3))
 
-  # mcmc_areas_ridges_data returns parameters in their original order.
-  # mcmc_areas_ridges() reverses these levels internally so that the first
-  # parameter is placed at the top of the plot without needing `limits =` in
-  # scale_y_discrete.
-  data <- mcmc_areas_ridges_data(vdiff_dframe, pars = pars3)
-  par_levels <- levels(data$parameter)
-  expect_equal(par_levels[1], "V1")
-  expect_equal(par_levels[length(par_levels)], "V3")
+  # mcmc_areas_ridges() reverses factor levels internally so that the first
+  # parameter sits at the top of the plot without relying on `limits =`.
+  # Check the built plot data to confirm the reversal.
+  plot_data <- ggplot2::ggplot_build(p_base)
+  y_limits <- plot_data$layout$panel_scales_y[[1]]$get_limits()
+  expect_equal(y_limits[1], "V3")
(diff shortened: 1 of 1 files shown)


jgabry:
Thanks for the PR! I'm not sure this is working correctly. This does fix the overlay issue when using `scale_y_discrete`, however when I try the examples from https://github.com/stan-dev/bayesplot/issues/287 using this branch it seems like the y axis labels are reordered but the actual density plots aren't reordered. Is that right? Do you see the same thing when you try those examples or is it working correctly for you?

utkarshpawade:
@jgabry  You're right, and thanks for catching that. After further testing, I can confirm the issue is not fully resolved. The original crash/overlay-breaking behaviour when adding scale_y_discrete() is fixed, but the densities do not correctly follow their labels when reordering or relabelling via scale_y_discrete() — the labels change but the density shapes stay in their original positions.

PR #438, opened 2026-03-10
Add unified wrapper functions for plot families

### LINKED ISSUE #433: Unified wrapper functions for bayesplot families
## Summary
Currently, each plot family in bayesplot exposes several closely related functions that share common inputs but differ in specific behaviors. For example, the error family within the PPC module includes:

+ `ppc_error_hist()`
+ `ppc_error_hist_grouped()`
+ `ppc_error_scatter()`
+ `ppc_error_binned()`
+ ...

Exposing all these individual functions in the documentation can make it difficult for users to navigate the package. To improve usability, I would propose providing a wrapper function for each family that consolidates these related functions into a single "general concept".

## Proposed Change
Introduce a high-level wrapper for each family that dispatches to the underlying functions via a type argument. Below is an example of the intended structure (subject to further refinement):

```
ppc_error <- function(
    y,
    yrep,
    type = c("hist", "scatter", "scatter_avg", "scatter_avg_vs_x", "binned"),
    group = NULL,
    x = NULL,
    grouped = FALSE,
    ...
    )
```
We could implement a similar pattern for other families. This approach would result in a more organized API and more compact documentation pages.

ishaan-arora-1:
Fixes #433

This adds high-level wrapper functions that let users access each plot family through a single function + `type` argument, instead of needing to know the exact function name upfront.

### New wrappers

**PPC:** `ppc_error()`, `ppc_dist()`, `ppc_discrete()`, `ppc_loo()`  
**PPD:** `ppd_dist()`  
**MCMC:** `mcmc_dist()`, `mcmc_trace_w()`, `mcmc_diag()`, `mcmc_nuts()`, `mcmc_recover()`

For example, instead of remembering `ppc_dens_overlay()` vs `ppc_ecdf_overlay()` vs `ppc_hist()`, users can call:

```r
ppc_dist(y, yrep, type = "dens_overlay")
ppc_dist(y, yrep, type = "hist")
ppc_dist(y, yrep, type = "ecdf_overlay", grouped = TRUE, group = g)
```

Partial matching on `type` is supported. Families with `_grouped` variants accept `grouped = TRUE`. All other arguments are forwarded via `...`.

The underlying functions remain fully available — these wrappers are thin dispatchers that just make the API a bit more discoverable.

### What's included

- `R/wrappers.R` — all wrapper implementations with a shared `.dispatch()` helper
- `man/bayesplot-wrappers.Rd` — consolidated docs listing valid types per family
- `tests/testthat/test-wrappers.R` — tests covering dispatch, grouped variants, error cases, and partial matching
- `NEWS.md` and `NAMESPACE` updated


jgabry:
Let's wait on this until we've had a chance to discuss in #433

jgabry:
In general for things like this that require discussion we should wait to open PRs until we've had the discussion and agreed on the new design. For things like fixing deprecations or fixing documentation (or other things like that that don't require discussion) it's ok to just open the PR.

ishaan-arora-1:
> In general for things like this that require discussion we should wait to open PRs until we've had the discussion and agreed on the new design. For things like fixing deprecations or fixing documentation (or other things like that that don't require discussion) it's ok to just open the PR. 

Alright, thankyou for the feedback!! Will keep in mind.

Issue #436, opened 2026-03-10
Expanding the bayesplot vignette library


florence-bockting:
## Summary
While bayesplot currently offers a few introductory vignettes, the user experience could be improved by more targeted documentation. In the following three suggestions:

### 1. "Which plot should I use?"
This problem-oriented vignette would provide a structured overview (e.g., a lookup table) mapping specific diagnostic questions to their corresponding plot functions.

| Diagnostic Question | Recommended Function(s) |
|-------------------------------|----------------------------------------|
| Did my MCMC chains mix? | `mcmc_trace()`, `mcmc_rank_overlay()` |
| Is my R-hat acceptable? | `mcmc_rhat()`, `mcmc_rhat_hist()` |
| ... | ... |

### 2. "Bayesplot and the Bayesian Workflow"
A workflow-oriented vignette that walks users through a complete modeling cycle using a single running example. It would demonstrate the appropriate plots for each stage.

### 3. "Customizing Bayesplot Outputs"
Since bayesplot returns standard ggplot objects, they can be modified using standard ggplot2 functions. However, it might be that some users are unaware of this extensibility. This vignette would guide users on how to benefit from this composability, covering topics such as: Adding layers or changing labels using the + operator, utilizing built-in color schemes and themes, customizing plot appearance via ggplot2 theme elements.

## Relation to other issues
This issue is related to [Issue#144](https://github.com/stan-dev/bayesplot/issues/144) (Extend and Update vignettes) and [Issue#142](https://github.com/stan-dev/bayesplot/issues/142) (Vignette/template demonstrating how to contribute a new plotting function to bayesplot).

jgabry:
I agree these would be good additions

ishaan-arora-1:
Hey @florence-bockting 

I did my research and looked at how arviz organised their docs, to get a better idea of how i should go about with this. 

I've been thinking about the questions from #440, which vignettes, what content, how they relate, long vs. focused.

Curently the three exitsing vignettes are organized by function families. They work good as reference docs, but don't help a user with quesitons like what should they check after they fit a model?

What i think is that on each vignette:

1. "Which plot should I use?" is useful, but overlaps somewhat with the visual gallery in #443. I think that a decision guide serves a different purpose than a gallery, but it could also work as a section within the workflow vignette rather than standalone. 
[...]

Issue #142, opened 2018-04-13
Vignette/template demonstrating how to contribute a new plotting function to bayesplot


jgabry:
Given the increasing number of people developing R packages depending on Stan, it would be nice to have a vignette demonstrating the process of contributing a new plotting function to **bayesplot**, including the coding (e.g., internally using the available `_data` functions or making a new one), doc, tests, pull request, etc? Something similar to @tjmahr's blog post but specifically focused on providing a template for new contributors. There are enough packages listing

PR #439, opened 2026-03-10
Add vignette sections for *_data() companion functions

ishaan-arora-1:
Fixes #435

Added vignette sections in the PPC and MCMC vignettes demonstrating how to use the `*_data()` companion functions for building custom ggplot2 visualizations, as requested in the issue.

AI Usage policy :
Parts of the documentation were generated using an llm

PR #440, opened 2026-03-10
Add three new vignettes to expand the vignette library

ishaan-arora-1:
## Summary

Addresses #436 by adding three new vignettes to improve discoverability and onboarding:

Also related to #144 (extend vignettes) and #142 (contributor template vignette).

AI Usage Policy :
Major portion of this documentation was generated using an LLM.

### DIFF
--- a/vignettes/bayesian-workflow.Rmd
+++ b/vignettes/bayesian-workflow.Rmd
@@ -1,0 +1,255 @@
+---
+title: "Bayesplot and the Bayesian workflow"
+author: "bayesplot team"
+date: "`r Sys.Date()`"
+output: 
+  rmarkdown::html_vignette:
+    toc: true
+    toc_depth: 3
+params:
+  EVAL: !r identical(Sys.getenv("NOT_CRAN"), "true")
+vignette: >
+  %\VignetteIndexEntry{Bayesplot and the Bayesian workflow}
+  %\VignetteEngine{knitr::rmarkdown}
+  %\VignetteEncoding{UTF-8}
+---
+
+```{r, child="children/SETTINGS-knitr.txt"}
+```
+```{r pkgs, include=FALSE}
+library("ggplot2")
+library("rstanarm")
+library("bayesplot")
+bayesplot_theme_set()
+set.seed(1234)
+```
+
+## Introduction
+
+A Bayesian analysis is more than fitting a single model and reading off the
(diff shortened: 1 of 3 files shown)


florence-bockting:
Thanks, @ishaan-arora-1, for the PR. I really appreciate your enthusiasm for tackling these problems. I realized my original issue description might not have been clear enough regarding its status. Before we dive into the solution, I think we should discuss a few things to make sure we're on the right track:

- Which vignettes do we want to have?
- What should be the content of each?
- What should be the relation between the vignettes?
- Do we want to have one long vignette or rather smaller, more focused vignettes?
- Etc.

So, there is some uncertainty here and we should first agree on a path that we want to go. Good ideas to move forward is for example to do first some research on resources about "How to write good documentation" and then we need to start thinking about the "structure/content of vignettes and relation between vignettes" such that they are most useful from the perspective of the user.

@jgabry would you agree?

(Additionally, the same hint I mentioned in [PR#439](https://github.com/stan-dev/bayesplot/pull/439#issuecomment-4053050940), regarding rendering vignettes locally, applies to this PR as well.)

ishaan-arora-1:
> Thanks, @ishaan-arora-1, for the PR. I really appreciate your enthusiasm for tackling these problems. I realized my original issue description might not have been clear enough regarding its status. Before we dive into the solution, I think we should discuss a few things to make sure we're on the right track:
> 
> * Which vignettes do we want to have?
> * What should be the content of each?
> * What should be the relation between the vignettes?
> * Do we want to have one long vignette or rather smaller, more focused vignettes?
> * Etc.
> 
> So, there is some uncertainty here and we should first agree on a path that we want to go. Good ideas to move forward is for example to do first some research on resources about "How to write good documentation" and then we need to start thinking about the "structure/content of vignettes and relation between vignettes" such that they are most useful from the perspective of the user.
> 
> @jgabry would you agree?
> 
> (Additionally, the same hint I mentioned in [PR#439](https://github.com/stan-dev/bayesplot/pull/439#issuecomment-4053050940), regarding rendering vignettes locally, applies to this PR as well.)

Youre absolutely right @florence-bockting 
I'll close this PR for now. Happy to contribute to the discussion on what vignettes would be most useful and how they should be organized, before implementing a solution to this.

jgabry:
Thanks to you both, I agree

PR #443, opened 2026-03-10
Add example gallery vignette with thumbnail plots (#437)

### LINKED ISSUE #437: Package cheatsheet with thumbnail plots or an example gallery
## Summary
Currently, the bayesplot ["Functions"](https://mc-stan.org/bayesplot/reference/index.html) overview provides only a structured list of function links. For better user orientation it would be helpful to provide something like a one-page cheatsheet or an example gallery that shows the functions with small rendered example plots. See for example how ArviZ handels this: https://python.arviz.org/projects/plots/en/latest/gallery/index.html.

In this context, it would also be helpful to get help in the documentation for specific plotting functions, without being forwarded to a documentation page that provides help for the entire "function family" (see for example https://mc-stan.org/bayesplot/reference/PPC-errors.html).

utkarshpawade:
Fixes #437

Adds a new vignette, `Example Gallery`, that shows small example plots for the main `bayesplot` functions. The gallery groups plots by category so users can quickly see what each function produces and navigate to its documentation. The layout is loosely inspired by the ArviZ plots gallery.

## Changes

- Added `vignettes/example-gallery.Rmd`
- ~60 example plots organized under `PPC`, `PPD`, `MCMC`, and `HMC/NUTS`
- Each thumbnail links to the corresponding reference page
- Uses built-in example data (`example_y_data()`, `example_mcmc_draws()`, etc.)
- Rendering guarded by `NOT_CRAN`, consistent with other vignettes
- Added the gallery to `_pkgdown.yml` navbar and articles


florence-bockting:
Thank you @utkarshpawade for your PR. Overall this goes definitely in the direction that I had in mind.
Is this actually the default theme? 

I think we have to think a bit more about the presentation. Now, it is basically again a long document through which the user has to scroll. I wonder whether we can implement something more dynamic. Probably a Vignette is actually not the right format. Also a short one-sentence description might be helpful as well.

I think these are points we need to discuss first before we jump to the actual implementation. I think it might make sense to go one step back and discuss our thoughts and ideas in the discussion, which is already ongoing in the corresponding [issue#437](https://github.com/stan-dev/bayesplot/issues/437). Only when we have an idea of how the implementation should look like, a PR makes sense.

utkarshpawade:
yes, you're right, let's first discuss the format and presentation before jumping to implementation.

PR #446, opened 2026-03-10
Remove dead modify_aes_() and fix deprecated size usage in helpers

### LINKED ISSUE #444: Remove dead code: modify_aes_() wrapping deprecated aes_()
`modify_aes_()` in `R/helpers-gg.R` (lines 31–33) wraps the deprecated `aes_()` function. It is never called anywhere in the codebase — the non-deprecated replacement `modify_aes()` (lines 38–40) exists and is used instead.

The dead function should be removed along with its roxygen comment block.

### LINKED ISSUE #445: Deprecated size usage in exported helpers and documentation examples
A few places still use `size` where ggplot2 3.4+ expects `linewidth`:

**Exported API:**
- `grid_lines(color, size)` in `R/bayesplot-helpers.R` accepts a `size` parameter and its `@param` documents it as `size`, but internally passes it as `linewidth` to `element_line()`. The parameter should be renamed to `linewidth` (with `size` kept as a deprecated alias for back-compat).
- The internal `grid_lines_y()` has the same issue.

**Documentation examples:**
- `R/mcmc-diagnostics.R` line 80: `hline_at(0.5, linetype = 2, size = 0.15, color = "gray")` — should use `linewidth` since `hline_at()` forwards `...` to `geom_hline()`.
- `R/mcmc-scatterplots.R` line 112: `geom_smooth(method = "lm", se = FALSE, color = "gray20", size = .75, linetype = 2)` — should use `linewidth`.

These produce deprecation warnings on ggplot2 >= 3.4.0.

AI Usage Policy :
This issue was found using an LLM

ishaan-arora-1:
Fixes #444, fixes #445

- Removed `modify_aes_()` which wrapped the deprecated `aes_()` and was never called anywhere in the codebase.
- Renamed the `size` parameter to `linewidth` in `grid_lines()` and `grid_lines_y()` to match ggplot2 3.4+ semantics (the value was already being passed as `linewidth` internally).
- Updated doc examples in `mcmc_acf_bar` and `mcmc_scatter` that used `size` where `linewidth` is expected.

AI Usage Policy :
This issue was found using an LLM, but was solved without one!!


R/bayesplot-helpers.R
```diff
@@ -446,19 +446,19 @@ plot_bg <- function(on = TRUE, ...) {
 
 #' @rdname bayesplot-helpers
 #' @export
-#' @param color,size Passed to [ggplot2::element_line()].
+#' @param color,linewidth Passed to [ggplot2::element_line()].
 #'
-grid_lines <- function(color = "gray50", size = 0.2) {
+grid_lines <- function(color = "gray50", linewidth = 0.2) {
```
jgabry: because this particular function is exported we can't just remove the `size` argument without potentially breaking user code. I think we need to keep `size` and deprecate it if we want to add `linewidth`.

R/bayesplot-helpers.R
```diff
@@ -446,19 +446,19 @@ plot_bg <- function(on = TRUE, ...) {
...
   theme(
-    panel.grid.major = element_line(color = color, linewidth = size),
-    panel.grid.minor = element_line(color = color, linewidth = size * 0.5)
+    panel.grid.major = element_line(color = color, linewidth = linewidth),
+    panel.grid.minor = element_line(color = color, linewidth = linewidth * 0.5)
   )
 }
 
-grid_lines_y <- function(color = "gray50", size = 0.2) {
+grid_lines_y <- function(color = "gray50", linewidth = 0.2) {
```
jgabry, line 467: This function is just internal so here it's ok to remove `size`.

jgabry:
Also there's a test that needs to be updated: 

https://github.com/stan-dev/bayesplot/actions/runs/22903567636/job/66472428428?pr=446#step:7:1277

R/bayesplot-helpers.R
ishaan-arora-1: yes, nice catch!

jgabry:
The failure on r-devel is unrelated (there is an issue with tiny differences in SVGs on r-devel). Merging now. Thanks!

Issue #449, opened 2026-03-10
Replace stat_function() with geom_function() and fix override.aes size for line legends


ishaan-arora-1:
A few remaining ggplot2 modernization items I found while working on the other deprecation fixes:

**1. `overlay_function()` wraps `stat_function()`**

`overlay_function()` in `bayesplot-helpers.R` is a thin wrapper around `ggplot2::stat_function()`. Since ggplot2 3.3.0, `geom_function()` is the recommended replacement. The switch is straightforward since `overlay_function()` just passes `...` through.

**2. `override.aes = list(size = ...)` for line-based legends**

`R/mcmc-traces.R` (line ~722) uses `size` in `override.aes` for a linetype legend key, where `linewidth` is the correct aesthetic since ggplot2 3.4.0.

(The similar case in `ppd-distributions.R` is already covered by #430.)

PR #450, opened 2026-03-10
Fix override.aes size to linewidth for linetype legend (#449)

ishaan-arora-1:
Fixes the `override.aes` in the `linetype` legend for `mcmc_trace()` divergence plots — changes `size` to `linewidth`, which is the correct aesthetic for line widths since ggplot2 3.4.0.

The `stat_function` -> `geom_function` change has been reverted per review feedback.

### DIFF
--- a/R/bayesplot-helpers.R
+++ b/R/bayesplot-helpers.R
@@ -21,7 +21,7 @@
 #'   [ggplot2::element_line()].
 #'
 #'   For `overlay_function`, `...` is passed to
-#'   [ggplot2::geom_function()].
+#'   [ggplot2::stat_function()].
 #'
 #' @return
 #' A **ggplot2** layer or [ggplot2::theme()] object that can be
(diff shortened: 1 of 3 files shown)


jgabry:
Thanks. this seems like a change we should make but we have a failing test: 

https://github.com/stan-dev/bayesplot/actions/runs/22910316067/job/66500925400?pr=450#step:7:1175

ishaan-arora-1:
@jgabry only ubuntu fails, we can merge now right?

jgabry:
> `stat_function()`, which has been superseded by `geom_function()` since ggplot2 3.3.0 

Actually, I have a question about this. I just checked the doc for `stat_function` and it's not deprecated and it doesn't say that `geom_function` should be used instead. Where do you see that it has been superseded? Maybe I didn't look in the right place.

ishaan-arora-1:
> `overlay_function()` was wrapping `stat_function()`, which has been superseded by `geom_function()` since ggplot2 3.3.0. Switched it over —

Hey @jgabry, to reframe the pr, instead of 'superseded' i should say 'recommended alternative' .
the ggplot2 3.3.2 states : 'A newly added geom_function() is now recommended to use in conjunction with/instead of stat_function()'

you should find it here exactly : https://ggplot2.tidyverse.org/news/index.html#ggplot2-332

PR #451, opened 2026-03-10
Infer annotation aesthetics from the active ggplot2 theme (#120)

### LINKED ISSUE #120: infer annotation aesthetics from global theme

ishaan-arora-1:
Fixes #120.

Reference lines like the vertical zero-line in `mcmc_intervals()` and the rhat=1 line in `mcmc_rhat()` had hardcoded colors (`"gray90"`, `"gray"`) that didn't adapt when users switched ggplot2 themes. As @tjmahr pointed out in the issue, this makes them invisible or look wrong on themes like `theme_grey()` or dark themes.

Added an internal `annotation_style()` helper that reads the active theme's gridline color and linewidth to derive sensible reference line aesthetics. When gridlines are blank (bayesplot's default theme), it falls back to the previous `gray90`/`0.5` defaults so existing plots look exactly the same.

Updated the four internal call sites in `mcmc-intervals.R` (3x `vline_0`) and `mcmc-diagnostics.R` (1x `vline_at`) to use the new helper.

### DIFF
--- a/tests/testthat/_snaps/mcmc-diagnostics/mcmc-neff-theme-gray.svg
+++ b/tests/testthat/_snaps/mcmc-diagnostics/mcmc-neff-theme-gray.svg
@@ -21,141 +21,141 @@
 <rect x='0.00' y='0.00' width='720.00' height='576.00' style='stroke-width: 1.07; stroke: #FFFFFF; fill: #FFFFFF;' />
 </g>
 <defs>
-  <clipPath id='cpNS40OHw2MjIuNDh8MjIuNzh8NTQxLjQx'>
-    <rect x='5.48' y='22.78' width='617.00' height='518.63' />
+  <clipPath id='cpNS40OHw2MjIuNDh8MjIuNzh8NTQxLjEz'>
+    <rect x='5.48' y='22.78' width='617.00' height='518.35' />
   </clipPath>
 </defs>
-<g clip-path='url(#cpNS40OHw2MjIuNDh8MjIuNzh8NTQxLjQx)'>
-<rect x='5.48' y='22.78' width='617.00' height='518.63' style='stroke-width: 1.07; stroke: none; fill: #EBEBEB;' />
-<polyline points='34.86,541.41 34.86,22.78 ' style='stroke-width: 0.53; stroke: #FFFFFF; stroke-linecap: butt;' />
-<polyline points='108.31,541.41 108.31,22.78 ' style='stroke-width: 0.53; stroke: #FFFFFF; stroke-linecap: butt;' />
-<polyline points='225.84,541.41 225.84,22.78 ' style='stroke-width: 0.53; stroke: #FFFFFF; stroke-linecap: butt;' />
-<polyline points='372.74,541.41 372.74,22.78 ' style='stroke-width: 0.53; stroke: #FFFFFF; stroke-linecap: butt;' />
-<polyline points='519.65,541.41 519.65,22.78 ' style='stroke-width: 0.53; stroke: #FFFFFF; stroke-linecap: butt;' />
-<polyline points='622.48,541.41 622.48,22.78 ' style='stroke-width: 0.53; stroke: #FFFFFF; stroke-linecap: butt;' />
-<polyline points='5.48,526.01 622.48,526.01 ' style='stroke-width: 1.07; stroke: #FFFFFF; stroke-linecap: butt;' />
-<polyline points='5.48,500.33 622.48,500.33 ' style='stroke-width: 1.07; stroke: #FFFFFF; stroke-linecap: butt;' />
-<polyline points='5.48,474.66 622.48,474.66 ' style='stroke-width: 1.07; stroke: #FFFFFF; stroke-linecap: butt;' />
-<polyline points='5.48,448.99 622.48,448.99 ' style='stroke-width: 1.07; stroke: #FFFFFF; stroke-linecap: butt;' />
-<polyline points='5.48,423.31 622.48,423.31 ' style='stroke-width: 1.07; stroke: #FFFFFF; stroke-linecap: butt;' />
-<polyline points='5.48,397.64 622.48,397.64 ' style='stroke-width: 1.07; stroke: #FFFFFF; stroke-linecap: butt;' />
-<polyline points='5.48,371.96 622.48,371.96 ' style='stroke-width: 1.07; stroke: #FFFFFF; stroke-linecap: butt;' />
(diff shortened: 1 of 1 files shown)


R/helpers-gg.R
```diff
@@ -10,8 +10,9 @@
 #' @examples
 #' # Draw a vertical line at zero (or do nothing)
 #' xs <- -2:2
+#' style <- annotation_style()
 #' maybe_vertical_line <- if (0 > min(xs) && 0 < max(xs)) {
-#'   vline_0(color = "gray90", linewidth = 0.5)
+#'   vline_0(color = style$color, linewidth = style$linewidth)
```
jgabry: I think this example can be left as it was, it's just showing how to use `geom_ignore()` so we don't need to complicate it by also calling `annotation_style()`. It doesn't really matter what the colors are here.

R/helpers-gg.R
```diff
@@ -68,6 +76,29 @@ force_x_axis_in_facets <- function() {
   )
 }
 
+# Derive annotation line aesthetics from the active theme's gridlines.
+# When the theme has visible gridlines, the reference line inherits their
+# color at twice the major gridline width. When gridlines are blank (e.g.
+# bayesplot's default theme), falls back to a light gray.
+annotation_style <- function() {
```
jgabry: Is there a way to test that this function does what it's supposed to? At a minimum, for the plots that it affects you could add visual tests using the vdiffr framework we already use so we can see what they look like.

R/helpers-gg.R
```diff
@@ -68,6 +76,29 @@ force_x_axis_in_facets <- function() {
...
+  thm <- bayesplot_theme_get()
+  grid <- calc_element("panel.grid.major", thm)
+  if (inherits(grid, "element_blank") || is.null(grid)) {
+    return(list(color = "gray90", linewidth = 0.5))
+  }
+  minor <- calc_element("panel.grid.minor", thm)
```
jgabry: This is checking panel.grid.major / panel.grid.minor, but every place the function is called in this PR draws a vertical reference line. For that, I think the relevant theme elements are panel.grid.major.x / panel.grid.minor.x, which control vertical grid lines. The way this is written now I think that a theme that customizes only the vertical grid lines can be misread as having no grid leading to falling back to gray90 / 0.5 when it shouldn't. (Definitely worth double checking this though, I could be wrong about this!)

R/mcmc-diagnostics.R
```diff
@@ -154,8 +154,9 @@ mcmc_rhat <- function(rhat, ..., size = NULL) {
       bayesplot_theme_get()
 
   if (min(data$value) < 1) {
+    ref_style <- annotation_style()
     graph <- graph +
-      vline_at(1, color = "gray", linewidth = 1)
+      vline_at(1, color = ref_style$color, linewidth = ref_style$linewidth)
```
jgabry, line 160: If the fallback values are hardcoded as `gray90` and `0.5` then the behavior here isn't preserved (`gray` and `1`). Maybe annotation_style should take the fallback values as an argument since the values used here are different from the values in mcmc-intervals.R?

R/mcmc-diagnostics.R
jgabry: If the goal is theme-aware annotation aesthetics, this plot is still only partially updated because the dashed break lines here are still hardcoded `"gray"`.

ishaan-arora-1:
Hey @jgabry how do the changes look now?

R/helpers-gg.R
```diff
@@ -68,6 +68,49 @@ force_x_axis_in_facets <- function() {
...
+# Because every current call-site draws a *vertical* reference line we first
+# inspect the axis-specific element (`panel.grid.major.x`) so that themes
+# which only customise vertical gridlines are handled correctly, then fall
+# back to the general `panel.grid.major`.
+#
+# @param fallback_color,fallback_linewidth Values returned when the resolved
+#   grid element is blank or NULL (i.e. the theme hides gridlines).  Different
+#   plots historically used different hardcoded values, so callers can preserve
+#   backward-compatible defaults.
+annotation_style <- function(fallback_color = "gray90",
```
jgabry, line 82: annotation_style() should also fall back when the gridline is effectively invisible (color is `NA` or zero linewidth), not just when it is element_blank()/NULL. Right now if color is NA (which is possible) or the linewidth is 0 those themes will produce invisible annotations.

tests/testthat/test-mcmc-diagnostics.R
```diff
@@ -191,3 +191,20 @@ test_that("mcmc_acf_bar renders correctly", {
...
+
+test_that("mcmc_rhat and mcmc_neff annotations inherit from theme gridlines", {
+  testthat::skip_on_cran()
+  testthat::skip_if_not_installed("vdiffr")
+  skip_on_r_oldrel()
+
+  bayesplot_theme_set(ggplot2::theme_gray())
+  on.exit(bayesplot_theme_set(), add = TRUE)
+
+  rhats <- seq(from = 1, to = 1.20, length.out = 10)
```
jgabry, line 203: The solid reference line at x = 1 is only added when min(data$value) < 1. With this test data, min = 1, so the solid rhat = 1 line never appears.

PR #454, opened 2026-03-11
fix: replace `<-` with `=` for named argument in `ecdf_intervals()` call

### LINKED ISSUE #453: Assignment-in-call Bug (`<-` instead of `=`)
### Description

In `R/mcmc-traces.R`, the call to `ecdf_intervals()` passes the `L` argument using the assignment operator `<-` inside the function call.

```r
lims <- ecdf_intervals(
  gamma = gamma,
  N = n_iter,
  K = if (is.null(K)) { n_iter } else { K },
  L <- n_chain
)
```
Using `<-` inside a function call:

- Performs an assignment in the calling environment.
- Passes the value positionally instead of as a named argument.

### Impact

- May introduce an unintended variable (`L`) in the calling environment.
- Relies on positional argument matching rather than explicit naming.
- Could lead to subtle issues if the function signature changes in the future.

utkarshpawade:
Fix #453

Fixes the use of `<-` (assignment operator) instead of `=` (named argument) when passing `L` to `ecdf_intervals()` in `R/mcmc-traces.R`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 # bayesplot (development version)
 
-* Fix assignment-in-call bug in `mcmc_trace_ucdf`/`mcmc_trace_rcdf`: `L <- n_chain` (which leaked `L` into the calling environment) corrected to `L = n_chain` (#).
+* Fix assignment-in-call bug in `mcmc_rank_ecdf()` (#).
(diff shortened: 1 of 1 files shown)


jgabry:
Thanks, merging now.

PR #470, opened 2026-03-13
Make diagnostic color matching explicit for neff

### LINKED ISSUE #469: match.arg Mismatch in diagnostic_color_scale()
**File:** [mcmc-diagnostics.R](file:///c:/GSOC/Stan/bayesplot/R/mcmc-diagnostics.R#L430-L432) 

```r
diagnostic_color_scale <- function(diagnostic = c("rhat", "neff_ratio"), ...) {
  diagnostic <- match.arg(diagnostic)
  # ...
}
```

But callers pass `"neff"`:
```r
scale_color_diagnostic <- function(diagnostic = c("rhat", "neff")) {
  d <- match.arg(diagnostic)
  diagnostic_color_scale(d, aesthetic = "color")  # passes "neff"
}
```

When `"neff"` is passed to `diagnostic_color_scale()`, `match.arg` will partial-match it to `"neff_ratio"`. This works incidentally due to partial matching, but it's fragile and confusing. The parameter names should be consistent.

utkarshpawade:
Fixes #469 

This PR removes reliance on partial matching when resolving diagnostic names for color/fill scales.

Previously the wrapper accepted `neff` while internal helpers expected `neff_ratio`, which worked only because of partial matching. This change makes the accepted names explicit.

Changes
- Accept `rhat`, `neff`, and `neff_ratio`
- Normalize `neff` to `neff_ratio` internally
- Add tests for both names

### DIFF
--- a/tests/testthat/test-mcmc-diagnostics.R
+++ b/tests/testthat/test-mcmc-diagnostics.R
@@ -79,12 +79,6 @@ test_that("'description' & 'rating' columns are correct (#176)", {
   expect_equal(df$description, expected_descriptions)
 })
 
-test_that("diagnostic color helpers use neff_ratio consistently", {
-  expect_no_error(scale_color_diagnostic("neff_ratio"))
-  expect_no_error(scale_fill_diagnostic("neff_ratio"))
-  expect_no_error(diagnostic_color_scale("neff_ratio", aesthetic = "color"))
-})
-
 test_that("mcmc_acf & mcmc_acf_bar return a ggplot object", {
   expect_gg(mcmc_acf(arr, pars = "beta[1]", regex_pars = "x\\:[2,5]"))
   expect_gg(mcmc_acf_bar(arr, pars = "beta[1]", regex_pars = "x\\:[2,5]"))


utkarshpawade:
Yes, that makes sense and it aligns with a minimal change approach; I’ll inspect the current branch state (including any recent edits) and then simplify the implementation exactly as requested.

tests/testthat/test-mcmc-diagnostics.R
```diff
@@ -79,6 +79,12 @@ test_that("'description' & 'rating' columns are correct (#176)", {
   expect_equal(df$description, expected_descriptions)
 })
 
+test_that("diagnostic color helpers use neff_ratio consistently", {
+  expect_no_error(scale_color_diagnostic("neff_ratio"))
+  expect_no_error(scale_fill_diagnostic("neff_ratio"))
+  expect_no_error(diagnostic_color_scale("neff_ratio", aesthetic = "color"))
+})
+
```
jgabry: I don't think these tests are necessary since we already essentially test for this error by testing the diagnostic plotting functions themselves. If `neff_ratio` caused an error then those tests would already fail.

PR #477, opened 2026-03-14
Replace superseded coord_flip() with swapped aesthetics (#476)

### LINKED ISSUE #476: Replace superseded coord_flip() with swapped aesthetics in ppc_error_scatter_avg
`ppc_error_scatter_avg()` and the deprecated `ppc_error_scatter_avg_vs_x()` in `R/ppc-errors.R` use `coord_flip()`, which is marked as [Superseded] in ggplot2.

From the ggplot2 docs:
> This function is superseded because in many cases, `coord_flip()` can easily be replaced by swapping the x and y aesthetics, or optionally setting the `orientation` argument in geom and stat layers.

https://ggplot2.tidyverse.org/reference/coord_flip.html

Two call sites at lines 260 and 330 of `R/ppc-errors.R`.

Would it be welcome to replace these with swapped aesthetics? Happy to open a PR if so.

ishaan-arora-1:
Fixes #476.

`coord_flip()` is marked `[Superseded]` in ggplot2 — replaced with direct aesthetic swapping in `ppc_error_scatter_avg()` and the deprecated `ppc_error_scatter_avg_vs_x()`.

Instead of building the plot with the axes one way and then flipping with `coord_flip()`, the x and y aesthetics are now mapped directly to the intended axes. Visual output is identical (all vdiffr snapshots pass unchanged).

https://ggplot2.tidyverse.org/reference/coord_flip.html

### DIFF
--- a/R/ppc-errors.R
+++ b/R/ppc-errors.R
@@ -243,22 +243,22 @@ ppc_error_scatter_avg <-
 
     stat <- as_tagged_function({{ stat }})
 
-    ppc_scatter_avg(
+    p <- ppc_scatter_avg(
       y = if (is_null(x)) y else x,
       yrep = errors,
       size = size,
       alpha = alpha,
       ref_line = FALSE,
       stat = stat
-    ) +
-      labs(
-        x = error_avg_label(stat),
-        y = if (is_null(x)) y_label() else as_label((qx))
-        ) + if (is_null(x)) {
-          NULL
-        } else {
-          coord_flip()
-        }
+    )
+
+    if (is_null(x)) {
+      p + labs(x = error_avg_label(stat), y = y_label())
+    } else {
+      p +
+        aes(x = .data$y_obs, y = .data$value) +
+        labs(x = as_label((qx)), y = error_avg_label(stat))
+    }
   }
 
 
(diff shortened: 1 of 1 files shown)


jgabry (commented):
I think we need to think about whether we want to do this. The snapshots staying the same only shows that the pictures are unchanged, but with this PR the plot object itself changes. By using remapped aesthetics it changes how user customization works after the plot is created, e.g. `scale_x_*`, `scale_y_*`, and axis theme tweaks. So this technically breaks backwards compatibility. 

That said, I can see an argument that this new behavior is a bit less confusing for users, since modifying a plot after coord_flip() can be awkward. So this PR is probably an improvement for _new_ users, but it can break _existing_ users' code if they have code that modifies these ppc-error plots. For that reason (not breaking existing code), I would lean towards keeping `coord_flip`, but I can also see the argument for changing it.

ishaan-arora-1:
> I think we need to think about whether we want to do this. The snapshots staying the same only shows that the pictures are unchanged, but with this PR the plot object itself changes. By using remapped aesthetics it changes how user customization works after the plot is created, e.g. `scale_x_*`, `scale_y_*`, and axis theme tweaks. So this technically breaks backwards compatibility.
> 
> That said, I can see an argument that this new behavior is a bit less confusing for users, since modifying a plot after coord_flip() can be awkward. So this PR is probably an improvement for _new_ users, but it can break _existing_ users' code if they have code that modifies these ppc-error plots. For that reason (not breaking existing code), I would lean towards keeping `coord_flip`, but I can also see the argument for changing it.

sure @jgabry, i get what youre saying 

That said, if the project eventually does a major version bump where breaking changes are expected, this would be a nice cleanup to bundle in i think

jgabry:
Yes I agree that if we make a major version bump this would be good to include.

PR #479, opened 2026-03-14
Replace deprecated reshape2::melt() with base R and tidyr equivalents

### LINKED ISSUE #478: Replace deprecated reshape2::melt() with tidyr::pivot_longer() and base R equivalents
### Problem

The package currently depends on `reshape2::melt()` in several internal helpers. The `reshape2` package has been superseded and is no longer actively maintained. Using it risks future compatibility issues if it is eventually archived on CRAN.

### Affected files
- `R/helpers-mcmc.R` — `melt_mcmc.mcmc_array`, `melt_mcmc.matrix`
- `R/bayesplot-extractors.R` — `log_posterior.stanfit`, `log_posterior.CmdStanMCMC`, `nuts_params.list`, `nuts_params.CmdStanMCMC`
- `R/helpers-ppc.R` — `melt_predictions`
- `R/ppc-discrete.R` — `.ppc_bars_data`
- `R/ppd-test-statistics.R` — `.ppd_stat_data`

utkarshpawade:
Fixes #478
## Summary

- Replace all 9 uses of `reshape2::melt()` across 5 source files with
  `base::as.data.frame.table()`, `tidyr::pivot_longer()`, and explicit
  `dplyr::bind_rows()` constructions
- Remove `reshape2` from `DESCRIPTION` Imports — it is no longer needed
- Update 10 vdiffr visual snapshots for `ppd_stat` plots affected by the migration
- Update inline documentation to remove references to `reshape2::melt()`

### DIFF
--- a/tests/testthat/_snaps/mcmc-diagnostics/mcmc-neff-default.svg
+++ b/tests/testthat/_snaps/mcmc-diagnostics/mcmc-neff-default.svg
@@ -20,106 +20,106 @@
 <g clip-path='url(#cpMC4wMHw3MjAuMDB8MC4wMHw1NzYuMDA=)'>
 </g>
 <defs>
-  <clipPath id='cpNS45OHw1OTYuODl8MjQuODV8NTM4LjI3'>
-    <rect x='5.98' y='24.85' width='590.91' height='513.42' />
+  <clipPath id='cpNS45OHw1OTYuODl8MjQuODV8NTM3Ljk2'>
+    <rect x='5.98' y='24.85' width='590.91' height='513.11' />
   </clipPath>
 </defs>
-<g clip-path='url(#cpNS45OHw1OTYuODl8MjQuODV8NTM4LjI3)'>
-<line x1='5.98' y1='523.02' x2='5.98' y2='523.02' style='stroke-width: 1.16; stroke: #011F4B; stroke-linecap: butt;' />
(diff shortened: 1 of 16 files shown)


utkarshpawade:
@jgabry The 10 uncovered lines are in R/bayesplot-extractors.R, specifically the log_posterior.CmdStanMCMC and nuts_params.CmdStanMCMC methods which require a CmdStan installation to test. These lines were already uncovered before this PR, I just replaced uncovered reshape2 code with equivalent uncovered base R code. No regression in coverage.

jgabry (changes requested):
A few issues: 

Several new factor(...) conversions reorder parameters alphabetically instead of preserving the original dimname/column order, which can silently change plot axis and facet ordering. 

Also, melt_mcmc.matrix() now produces invalid Draw values for matrices without row names because as.data.frame.table() creates synthetic labels that are then coerced to NA. 

So these would at least need to be fixed if we want to make these changes. However, reshape2 is superseded but still likely to remain on CRAN. I would probably prefer not to replace it unless the replacement is just as simple or equally clear. I think as is this changes too much code for it to be worth it unless reshape2 is going to be removed from CRAN. So I would lead towards closing this unless there's a way to make these changes much more simply.

utkarshpawade:
Closing this. reshape2 looks stable on CRAN and fixing the factor ordering + NA Draw regressions would only make the code more complex. Not worth it unless reshape2 actually gets removed. Thanks for the review!

PR #481, opened 2026-03-15
Fix melt_mcmc.mcmc_array() silently ignoring the as.is argument

### LINKED ISSUE #480: melt_mcmc.mcmc_array() ignores the as.is argument
The `as.is` parameter in `melt_mcmc.mcmc_array()` is never actually used. The function accepts it in the signature but internally always passes `as.is = FALSE` to `reshape2::melt()`, so whatever the caller passes gets silently thrown away.

This matters for anyone relying on character/factor conversion behavior in the melted output, they'd assume `as.is = FALSE` does something, but it doesn't.

As Per the https://www.rdocumentation.org/packages/reshape2/topics/melt.array,
`as.is` controls whether character columns are converted to factors in the
melted output. The current behavior makes that control inaccessible to callers.

utkarshpawade:
Fixes #480 
`as.is` was hardcoded to `FALSE` inside the function regardless of what the caller passed. Changed it to forward the argument correctly. Default behavior is unchanged since the signature already defaults to `TRUE`, and the `Parameter` column is still explicitly converted to a factor after melting
(see `R/helpers-mcmc.R` line 144), so that behavior is unaffected by this fix.

Reference: https://www.rdocumentation.org/packages/reshape2/topics/melt.array

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 # bayesplot (development version)
 
-* Fix `melt_mcmc.mcmc_array()` ignoring the `as.is` argument — the value passed by the caller is now forwarded correctly to `reshape2::melt()` instead of being silently overridden with `FALSE`.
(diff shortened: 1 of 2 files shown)


utkarshpawade:
Closing this - after digging into it, melt_mcmc is an internal unexported function so the unused as.is parameter has no impact on any public API or downstream behavior. Not worth the churn.

PR #485, opened 2026-03-15
Replace c(mult, add) with expansion() in scale expand args

### LINKED ISSUE #484: Standardize scale expand arguments to use expansion() helper instead of c(mult, add)
The codebase uses `expand = c(mult, add)` in several places while `mcmc-intervals.R` and `ppc-loo.R` already use the `expansion()` helper introduced in ggplot2 3.3.0. Would be good to standardize everything to use `expansion()` for consistency.

From the ggplot2 3.3.0 changelog:
"The expansion() helper function has been added to make it easier to specify the expand argument in scale_*() functions."

Affected files: `helpers-gg.R`, `mcmc-diagnostics.R`, `mcmc-diagnostics-nuts.R`, `mcmc-distributions.R`, `mcmc-parcoord.R`

utkarshpawade:
Fixes #484 

Summary
•Replaced all old-style expand = c(mult, add) numeric vector arguments with the expansion() helper across helpers-gg.R, mcmc-•diagnostics.R, mcmc-diagnostics-nuts.R, mcmc-distributions.R, and mcmc-parcoord.R
•The expansion() helper was introduced in ggplot2 3.3.0 and is the documented modern approach; the old c(mult, add) format is still supported but discouraged
•Brings these files in line with the style already used in mcmc-intervals.R and ppc-loo.R.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -12,6 +12,7 @@
 * Default to `quantiles=100` for all dot plots by @behramulukir (#402)
 * Use `"neff_ratio"` consistently in diagnostic color scale helpers to avoid relying on partial matching of `"neff"`.
 * Replace `expand = c(mult, add)` with `ggplot2::expansion()` helper in scale functions for consistency with ggplot2 >= 3.3.0 style.
+* Replace uses of `geom_bar(stat = "identity")` with the more idiomatic ggplot2 form `geom_col()` 
 
 # bayesplot 1.15.0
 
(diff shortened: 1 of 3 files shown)


R/helpers-gg.R
```diff
@@ -37,10 +37,10 @@ modify_aes <- function(mapping, ...) {
 facet_wrap_parsed <- function(...) {
   facet_wrap(..., labeller = label_parsed)
 }
-dont_expand_y_axis <- function(expand = c(0,0)) {
+dont_expand_y_axis <- function(expand = expansion(0, 0)) {
```
jgabry, line 40: Changing the default here is good, but there are still places in the code where we call e.g.

dont_expand_y_axis(c(0.005, 0))

in which case this is still using the old behavior.

PR #487, opened 2026-03-15
Replace select(-c(col)) with select(-all_of(col))

### LINKED ISSUE #486: Replace select(-c("col")) with select(-all_of("col")) in mcmc-intervals.R
`mcmc-intervals.R` uses `select(-c("col"))` for deselection in two places (lines 759, 845), which is deprecated in dplyr 1.0.0+. The correct approach is to use `all_of()`:

```r
# deprecated
select(-c("center"), "m")

# correct
select(-all_of("center"), all_of("m"))

utkarshpawade:
Fixes #486 
**PR Description:**

- Replaced `select(-c("center"), "m")` and `select(-c("density"))` with `select(-all_of("center"), all_of("m"))` and `select(-all_of("density"))` in `mcmc-intervals.R` (lines 759, 845)
- Fixes deprecation warnings from dplyr 1.0.0+ when using character vectors inside `c()` for deselection in `select()`

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -11,6 +11,7 @@
 * New functions `mcmc_dots` and `mcmc_dots_by_chain` for dot plots of MCMC draws by @behramulukir (#402)
 * Default to `quantiles=100` for all dot plots by @behramulukir (#402)
 * Use `"neff_ratio"` consistently in diagnostic color scale helpers to avoid relying on partial matching of `"neff"`.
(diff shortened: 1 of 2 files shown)


jgabry:
> * Fixes deprecation warnings from dplyr 1.0.0+ when using character vectors inside `c()` for deselection in `select()`

I think these changes are good, but are you sure there are warnings? I don't get any deprecation warnings. Can you share the warning messages you were seeing?

utkarshpawade:
My bad, you're right. The warning is for external character variables, not inline strings like these. No actual warnings here, it's just a style thing. Happy to merge or close, up to you.

PR #489, opened 2026-03-15
Replace geom_smooth(stat = "identity") with geom_line() in ppc-discrete.R

### LINKED ISSUE #488: Replace geom_smooth(stat = "identity") with geom_line() in ppc-discrete.R
In `R/ppc-discrete.R` (line 322), `geom_smooth()` is used with `stat = "identity"`, which bypasses smoothing entirely and just connects pre-computed points — making it functionally identical to `geom_line()`. This is semantically misleading since a reader would expect a smoothing curve from `geom_smooth()`.

From the [ggplot2 docs on `geom_smooth()`](https://ggplot2.tidyverse.org/reference/geom_smooth.html): the default stat is `stat_smooth`. Setting `stat = "identity"` bypasses this, making `geom_smooth()` functionally equivalent to `geom_line()` — a confusing choice of geom.

Should be replaced with `geom_line()` which accurately reflects the intent.

utkarshpawade:
Fixes #488
## Summary
- Replaced `geom_smooth(stat = "identity")` with `geom_line()` in `R/ppc-discrete.R` (line 322)
- `geom_smooth()` with `stat = "identity"` bypasses smoothing and is functionally equivalent to `geom_line()`, but misleads readers into expecting a smoothing curve. See https://ggplot2.tidyverse.org/reference/geom_smooth.html
- Also removed the unused `fill` argument which has no effect on `geom_line()`


utkarshpawade:
My bad, I missed that it was also inheriting ymin/ymax for the ribbon, so yeah not equivalent at all. And you're right, it's internal code so the readability argument doesn't really hold. Closing this, makes sense to keep the current version.

jgabry:
> My bad, I missed that it was also inheriting ymin/ymax for the ribbon, so yeah not equivalent at all. 

No worries, these things can be pretty subtle sometimes.

PR #495, opened 2026-03-17
Warn when neff ratios exceed 1

### LINKED ISSUE #494: validate_neff_ratio() silently accepts values greater than 1
`validate_neff_ratio()` in `R/mcmc-diagnostics.R` checks that values are positive but does not flag values exceeding 1. While neff ratios > 1 can technically occur with certain ESS estimators, they are unusual and may indicate a problem with the model or sampler. Users currently get no feedback when this happens.

A warning would help users catch potential issues without breaking existing workflows.

utkarshpawade:
Fixes #494 
## Summary
- `validate_neff_ratio()` now issues an `rlang::warn()` when any values are greater than 1
- Updated existing tests and added a new `expect_warning()` assertion

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,6 @@
 # bayesplot (development version)
 
+* `mcmc_neff()` and `mcmc_neff_hist()` now warn when any neff ratios are greater than 1, as this is unusual and may indicate a problem with the model or sampler.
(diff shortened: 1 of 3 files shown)


jgabry:
With Stan when ESS (n_eff) exceeds the posterior sample size it is actually often a favorable sign, not a bad one, so I don't actually  think we need a warning here.

PR #517, opened 2026-03-28
is_chain_list: return FALSE for empty lists 

### LINKED ISSUE #516: Replace sapply() with vapply() for type safety
## Description

Nine uses of `sapply()` can silently return a `list()` instead of the expected atomic vector on zero-length input, causing hard-to-diagnose downstream failures. `vapply()` enforces a declared return type and is faster due to pre-allocation.

The two `sapply(x, colnames)` calls at `helpers-mcmc.R:277,303` should be intentionally kept as they rely on `sapply()`'s flexible return to detect mismatched column names across chains.

utkarshpawade:
Fixes #516 
## Summary

- Replaces 9 uses of `sapply()` with `vapply()` across 5 files
- `vapply()` enforces a declared return type, preventing silent `list()` returns on zero-length input that can cause hard to diagnose downstream failures
- Two `sapply(x, colnames)` calls at `helpers-mcmc.R:277,303` are intentionally unchanged, they rely on flexible return types to detect mismatched column names across chains

### DIFF
--- a/R/helpers-mcmc.R
+++ b/R/helpers-mcmc.R
@@ -251,7 +251,7 @@ is_chain_list <- function(x) {
     return(FALSE)
   }
   check1 <- !is.data.frame(x) && is.list(x)
-  dims <- try(vapply(x, function(chain) length(dim(chain)), integer(1)), silent=TRUE)
+  dims <- try(sapply(x, function(chain) length(dim(chain))), silent=TRUE)
   if (inherits(dims, "try-error")) {
     return(FALSE)
   }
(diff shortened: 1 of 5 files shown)


R/helpers-mcmc.R
```diff
@@ -248,7 +248,7 @@ df_with_chain2array <- function(x) {
 #' @return TRUE or FALSE
 is_chain_list <- function(x) {
```
jgabry, line 249: Even with this PR `is_chain_list(list())` returns `TRUE` so we don't actually get a fix for empty chain lists. I think instead of `vapply()` we should just explicitly reject empty lists (error if `length(x) == 0`).

jgabry:
I think I accidentally reverted your changes, so I just pushed again to restore your changes. I think it's ready now

PR #519, opened 2026-03-28
Add unit test for .psis_subset() output structure

### LINKED ISSUE #518: Missing unit test for .psis_subset() output structure
`.psis_subset()` manually manipulates `psis` object internals because `loo` doesn't expose a public subset method (https://github.com/stan-dev/bayesplot/blob/master/R/ppc-loo.R#L722-L738). The existing tests only check that plotting works with a subset and that oversized subsets error, nothing verifies the returned object actually has the correct structure.

If `loo` changes its `psis` internals, this will break silently.

We should add a test that checks `.psis_subset()` preserves class, dimensions, field values, and attributes.

utkarshpawade:
Fixes #518 
## Summary
- Adds a test verifying `.psis_subset()` preserves psis object structure (class, log_weights, diagnostics, dims, norm_const_log, tail_len, r_eff)
- Acts as a canary if `loo` changes its psis internals

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 # bayesplot (development version)
 
-* Added unit test for `.psis_subset()` .
+* Added unit test for `.psis_subset()`.
 * Fixed `is_chain_list()` to correctly reject empty lists instead of silently returning `TRUE`.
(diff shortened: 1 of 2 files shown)


jgabry (commented):
I don't think we need this because we already have  ppc_loo_intervals and ppc_loo_ribbon tests that use the `subset` argument. Those tests would fail if loo changes something that breaks .psis_subset.

PR #521, opened 2026-03-28
Add singleton-dimension edge-case tests for exported `_data()` functions

### LINKED ISSUE #520: Missing dedicated edge-case tests for exported _data() functions
The exported `_data()` functions are user-facing data preparation helpers but lack dedicated tests for edge cases like zero-length inputs, single observation, and single draw. Existing tests only exercise these functions implicitly through their plotting counterparts.

**Functions:** `ppc_bars_data()`, `ppc_error_data()`, `ppc_scatter_data()`, `ppc_scatter_avg_data()`, `ppc_loo_pit_data()`, `ppd_data()`, `ppd_stat_data()`, `ppd_intervals_data()`

**Missing coverage:**
- Zero-length inputs
- Single observation / single draw
- Alternative code paths (e.g., `ppc_bars_data()` with `prob = 0`, `ppc_loo_pit_data()` with custom `bw`)

utkarshpawade:
Fixes #520 

## Summary

Adds singleton-dimension edge-case tests for exported _data() functions


jgabry (changes requested):
Thanks. A bunch of comments: 

* I don't think we should have an entirely separate test file just for edge cases. I would put tests for these functions together with the existing tests for these functions. 

* Regarding the singleton-dimension tests: these are good and they're worth keeping but they should go with the other tests for these functions not in a separate file. 

* Regarding the tests for empty inputs: the exported `_data()` helpers do not currently have a consistent empty (zero length) input behavior: some return empty tibbles, `ppc_bars_data()` errors, and `ppd_stat_data()` currently fails with a low-level row-mismatch error. The tests here enforce this inconsistent behavior. I think we should first standardize the behavior (we'd need to decide on what it should be, that could be handled separately from this PR, we could discuss in an issue) and only then add tests. I'm glad you added these tests even though I don't want to keep them because it made me realize that this behavior isn't standardized across the functions! 

* Regarding the `ppc_loo_pit_data()` additions:  this feels over-specified relative to the value it adds. We already have  coverage for both `boundary_correction` paths and the returned structure. I’d keep one focused test for the `pit =` shortcut path and remove the repeated `expect_message(..., "pit")` / `is_y_label` checks for now.

tests/testthat/test-ppc-test-statistics.R
```diff
@@ -129,6 +129,16 @@ test_that("ppc_stat_data and ppd_stat_data throw correct errors", {
...
 })
 
+test_that("ppd_stat_data handles single draw and single observation", {
+  yrep_single <- matrix(rnorm(10), nrow = 1)
+  d <- ppd_stat_data(yrep_single, stat = "mean")
+  expect_equal(nrow(d), 1)
+
+  yrep_1obs <- matrix(rnorm(5), ncol = 1)
+  d2 <- ppd_stat_data(yrep_1obs, stat = "mean")
+  expect_s3_class(d2, "data.frame")
```
jgabry, line 139: Can you check `nrow(d2)` like you already check `nrow(d)` above?

jgabry (changes requested):
Thanks, looks good, just one small change to make

jgabry (approved):
Thank you!

PR #524, opened 2026-04-02
Add tests for theme/aesthetic helper functions

### LINKED ISSUE #523: Add dedicated tests for theme/aesthetic helper functions
The exported helpers `plot_bg()`, `panel_bg()`, `facet_bg()`, `legend_move()`, `legend_text()`, and `overlay_function()` appear in `test-convenience-functions.R` but coverage is minimal. tests only check parameter forwarding via `expect_identical()`.

No tests currently verify:

- **Return types** : background helpers return incomplete `theme` objects; `overlay_function()` returns a `LayerInstance`
- **`on = FALSE`** : that the correct element is set to `element_blank`
- **Composition** : adding helpers to bayesplot plots (e.g. `ppc_hist(...) + plot_bg(fill = "gray")`) doesn't error
- **Edge cases** : `legend_move("none")` behaving like `legend_none()`

utkarshpawade:
Fixes #523 

- Added `legend_move("none")` equivalence with `legend_none()`


jgabry (changes requested):
Thanks @utkarshpawade. I think the legend_move("none") test is worth keeping, since that behavior is explicitly documented and it guards a real contract of the helper.

For the rest, I’m not convinced the added tests are giving us much benefit. Most of these helpers are very thin wrappers around ggplot2::theme() or ggplot2::stat_function(), and the existing tests already cover the main behavior by checking the returned objects directly. The new composition tests mostly confirm that ggplot can still add a valid theme/layer object to a plot, which is more ggplot behavior than bayesplot behavior. The extra on = FALSE checks also seem redundant with the existing expectations that already compare against element_blank() themes.

So my preference would be to keep the legend_move("none") test and drop the others, to avoid adding test volume without much additional protection.

jgabry (approved):
Thanks, looks good! I'm looking into the failing CI run. It's not related to this PR, but I might try to fix it before merging more PRs.

PR #526, opened 2026-04-06
Replace apply() with storage.mode() for integer to numeric conversion

### LINKED ISSUE #525: Replace inefficient apply() with storage.mode() for integer-to-numeric matrix conversion
in `helpers-ppc.R`, `validate_predictions()` converts integer matrices to numeric using `apply(predictions, 2, as.numeric)`, which iterates column by column and creates an intermediate matrix. There is also a special-case branch for single-row matrices

`storage.mode(predictions) <- "numeric"` performs the same conversion in place without intermediates and works for all matrix dimensions, making the `nrow == 1` branch unnecessary

utkarshpawade:
Fixes #525

replace `apply(predictions, 2, as.numeric)` in `validate_predictions()` with `storage.mode(predictions) <- "numeric"`, an in place conversion that avoids column wise iteration and intermediates. this also removes the unnecessary `nrow == 1` special case since `storage.mode<-` handles all matrix dimensions uniformly

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,6 @@
 # bayesplot (development version)
 
+* Replace `apply()` with `storage.mode()` for integer-to-numeric matrix conversion in `validate_predictions()`.
 * Fixed `is_chain_list()` to correctly reject empty lists instead of silently returning `TRUE`.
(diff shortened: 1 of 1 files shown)


jgabry (approved):
Thanks!

PR #529, opened 2026-04-07
Fix validate_chain_list() colnames check to compare all chains

### LINKED ISSUE #528: validate_chain_list() only compares colnames of chains 1 and 2, ignoring chains 3+
In `validate_chain_list()` (`R/helpers-mcmc.R`), the colnames consistency check uses `identical(cnames[, 1], cnames[, 2])`, which only compares the first two chains. If chain 3 or later has different parameter names/order, the validation silently passes.

utkarshpawade:
Fixes #528 
- Fixed bug where `identical(cnames[, 1], cnames[, 2])` only validated colnames of chains 1 and 2, silently ignoring mismatches in chains 3+
- Replaced `sapply` with `lapply` + `vapply` to avoid matrix/list type ambiguity and compare all chains against the first


jgabry (approved):
Thanks, this is a good fix and the code is even cleaner than the original.

PR #533, opened 2026-04-08
Use geom_step() as default in ppc_ecdf_overlay() and deprecate discrete argument

### LINKED ISSUE #259: ppc_ecdf_overlay(): geom = "step"
In #234, we discussed the following:

> * I tried to stick as close as possible to `ppc_ecdf_overlay()`, with one exception: In `ppc_km_overlay()`, I hard-coded the use of `geom_step()`. In contrast, `ppc_ecdf_overlay()` has an argument `discrete` which defaults to `FALSE` so that `geom = "line"` is used. I am using `geom_step()` in `ppc_km_overlay()` because in my opinion, the ECDF (and therefore also the ECCDF and the Kaplan-Meier estimate) is *defined* to be a step function. If you want, I can open a new issue for this as far as `ppc_ecdf_overlay()` is concerned.

I somehow lost sight of this. Have you thought about hard-coding the use of `geom = "step"` in `ppc_ecdf_overlay()` (and `ppc_ecdf_overlay_grouped()`) or setting the default of argument `discrete` to `TRUE`?

utkarshpawade:
Fixes #259

The ECDF is a step function by definition, regardless of whether the data is continuous or discrete. `ppc_km_overlay()` already hard-codes `geom_step()` for the same reason. This PR makes `ppc_ecdf_overlay()` consistent.

## Changes

- `ppc_ecdf_overlay()`, `ppc_ecdf_overlay_grouped()`, and `ppd_ecdf_overlay()` now always use `geom = "step"` in `stat_ecdf()`
- The `discrete` argument is deprecated via `lifecycle::deprecate_warn()`
- Updated roxygen documentation and examples
- Regenerated vdiffr snapshots
- Added deprecation and no-warning tests

before
after

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 # bayesplot (development version)
 
-* `ppc_ecdf_overlay()`, `ppc_ecdf_overlay_grouped()`, and `ppd_ecdf_overlay()` now always use `geom_step()`. The `discrete` argument is deprecated .
+* `ppc_ecdf_overlay()`, `ppc_ecdf_overlay_grouped()`, and `ppd_ecdf_overlay()` now always use `geom_step()`. The `discrete` argument is deprecated.
 * Replace `apply()` with `storage.mode()` for integer-to-numeric matrix conversion in `validate_predictions()`.
 * Fixed `is_chain_list()` to correctly reject empty lists instead of silently returning `TRUE`.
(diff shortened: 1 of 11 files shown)


jgabry (approved):
Thanks, I agree it makes sense to just always use geom_step

PR #537, opened 2026-04-10
Fix legend border artifact in ppc_freqpoly

### LINKED ISSUE #134: legend border on ppc_freqpoly_grouped
Not sure what's going on here.

```r
library(bayesplot)
y <- example_y_data()
yrep <- example_yrep_draws()
group <- example_group_data()
ppc_freqpoly_grouped(y, yrep[1:3,], group, alpha = .5)
```

image

utkarshpawade:
Fixes #134 
- `ppc_freqpoly()` and `ppc_freqpoly_grouped()` displayed a visible dark border around the `y` legend key due to the `color` aesthetic from `geom_area()` bleeding into the legend glyph
- Added `guides(fill = guide_legend(override.aes = list(color = NA)))` to suppress the outline color in legend keys

Before
image

After
image


R/ppc-distributions.R
```diff
@@ -468,6 +468,7 @@ ppc_freqpoly <-
       ) +
       scale_fill_ppc() +
       scale_color_ppc() +
+      guides(fill = guide_legend(override.aes = list(color = NA, alpha = 1))) +
```
jgabry: I think we only need color=NA but not alpha = 1. If the user customizes the alpha level then it's ok if the legend uses that alpha level, we don't need to force it to 1.

jgabry (changes requested):
Thanks! I made one review comment. Also, do we have the same issue with `ppd_freqpoly` or just the `ppc_` version?

utkarshpawade:
No, ppd_freqpoly doesn't have the issue, it calls legend_none() ([ppd-distributions.R:283](https://github.com/stan-dev/bayesplot/blob/44c3e5c67c30102465769f387e141ad1e8335965/R/ppd-distributions.R#L283)) since there's no y vs yrep distinction to show. The legend-border artifact is specific to ppc_freqpoly/ppc_freqpoly_grouped.

jgabry:
Thinking about this a bit more, maybe we actually want the border in the plot? For example, with low alpha levels y and yrep can look similar except for the outline. This PR would remove the outline. What do you think?

utkarshpawade:
I checked this and guide_legend(override.aes = list(color = NA)) only affects the legend key glyph, not the plot geoms. The geom_area() outlines in the plot are untouched, so the low-alpha distinguishability between y and yrep you mentioned is preserved.
Comparison at alpha = 0.3:
#### Before: 
legend key has a dark border box around the y glyph:
befe
#### After: 
legend key is clean; plot polygons still have their outlines.
afe

jgabry:
Ok thanks. However, now I'm also wondering why we would actually want to remove the border in the legend. The border is there because the plot also has a border. So isn't the legend more accurate if it includes the border?

utkarshpawade:
yes, you're right Closing this PR as working-as-intended.

jgabry:
Ok thanks. Also see here if you haven't seen it yet:  https://github.com/stan-dev/bayesplot/pull/545

PR #539, opened 2026-04-13
Allow NAs in prepare_mcmc_array

### LINKED ISSUE #250: Requiring no NAs in prepare_mcmc_array
When running the plotting functions (e.g mcmc_trace) on a cmdstan draws() object, the draws() object gets converted to an array using prepare_mcmc_array. This function currently have a requirement  built in testing for anyNA(x), and throws an error if there's any NA discovered. This causes an issue with my current model, which has parameter estimates that are of work-around "ragged array" format, where different columns have different numbers of observations, and the remaining cells are filled with NA to create a rectangular matrix.  It means I can't use the mcmc_trace without creating a new implementation of the function, removing the anyNA check. 

It's not uncommon for data structures to have intentional NA's in them, so I would suggest an error for this condition is too strict.

utkarshpawade:
Fixes #250 
`prepare_mcmc_array()` previously aborted on any `NA` in the input, blocking plotting workflows where parameters are stored as ragged arrays padded with `NA` (common with cmdstanr draws). The hard error is demoted to a warning so plots like `mcmc_trace`, `mcmc_dens`, etc. work directly on such inputs. Stricter `NA` checks elsewhere (`helpers-ppc.R` validators, `drop_NAs_and_warn` in diagnostics) are left intact, those guard summary computations where silently dropping `NA`s would mislead.
## Changes
- `R/helpers-mcmc.R`: `abort()` → `warn()` on `anyNA(x)`.
- `tests/testthat/test-helpers-mcmc.R`: updated test to assert the warning and successful return.


jgabry:
Thanks, I'm not necessarily opposed to this change, but I'd be curious to see an actual example where this is useful.

utkarshpawade:
```r
set.seed(1)
iter <- 500; chains <- 4
draws <- array(
  rnorm(iter * chains * 6),
  dim = c(iter, chains, 6),
  dimnames = list(NULL, NULL, c("theta[1,1]","theta[1,2]","theta[1,3]",
                                "theta[2,1]","theta[2,2]","theta[2,3]"))
)
# group 2 has fewer observations, its third column is NA-padded
draws[, , "theta[2,3]"] <- NA

mcmc_trace(draws, pars = c("theta[1,3]", "theta[2,3]"))
```
On master this errors with `NAs not allowed in 'x'..` With the changes it renders, valid parameter traces normally, NA-only parameter is an empty panel:
image

R/helpers-mcmc.R
```diff
@@ -29,7 +29,9 @@ prepare_mcmc_array <- function(x,
     abort("Arrays should have 2 or 3 dimensions. See help('MCMC-overview').")
   }
   if (anyNA(x)) {
-    abort("NAs not allowed in 'x'.")
+    warn(
+      "NAs were found in 'x'. `prepare_mcmc_array()` does not remove them; some plots may render with missing values dropped, while summary functions (e.g. intervals, densities, diagnostics) may produce misleading results or error. Consider removing NAs before plotting or summarizing."
+    )
```
jgabry, line 34: I think we should shorten this wall of text to something like: 

"NAs found in 'x'. These are passed through as-is and may affect the resulting plots."

jgabry (changes requested):
Thanks, this is good. I made one suggestion about the warning message. Also, can you add a test for downstream plotting behavior? I think it would be helpful to add a vdiffr visual test for `mcmc_trace` using something similar to the example you gave. I would test two things: 

- the version you had where one plot is empty (e.g. `draws[, , "theta[2,3]"] <- NA`)
- a version where only some rows are NA (e.g. `draws[10:100, , "theta[2,3]"] <- NA`)

In the first case there should be a missing plot, in the second case there should be a missing chunk of the trace plot but the rest should be plotted.

jgabry (approved):
Thanks!

PR #543, opened 2026-04-15
Accept posterior::draws objects in PPC/PPD functions

### LINKED ISSUE #542: PPC/PPD functions don't accept posterior::draws objects as yrep/ypred
While working on my GSoC prior vs posterior prototype (which accepts draws objects for both prior and posterior inputs), I noticed that `mcmc_*` functions handle `posterior::draws` objects fine — `prepare_mcmc_array()` checks `posterior::is_draws(x)` and coerces to array — but the PPC/PPD side doesn't.

`validate_predictions()` in `helpers-ppc.R` does:

```r
stopifnot(is.matrix(predictions), is.numeric(predictions))
```

So passing a `draws_matrix` or `draws_array` as `yrep` to any `ppc_*` function fails immediately, even though these are standard outputs from cmdstanr and brms workflows. Users have to manually call `as.matrix()` first.

Seems like adding a draws check at the top of `validate_predictions()` (similar to what `prepare_mcmc_array()` already does) would fix this for all PPC/PPD functions at once. Happy to put up a PR if that sounds reasonable.

ishaan-arora-1:
Closes #542

## Summary
- `validate_predictions()` now checks for `posterior::draws` objects and coerces them to a plain matrix before the existing validation logic runs
- This mirrors what `prepare_mcmc_array()` already does on the MCMC side, so PPC/PPD functions now accept the same input formats as MCMC functions
- All existing `ppc_*` and `ppd_*` functions benefit since they all go through `validate_predictions()`

## Changes
- `R/helpers-ppc.R`: added a `posterior::is_draws()` check at the top of `validate_predictions()`, coercing via `as_draws_matrix()` + `unclass()`
- `tests/testthat/test-helpers-ppc.R`: added tests for `draws_matrix`, `draws_array`, `draws_df`, and `draws_rvars` inputs, plus end-to-end tests with `ppc_dens_overlay`, `ppc_hist`, and `ppc_stat`

## Test plan
- [x] All existing PPC tests pass (510 pass, 0 fail)
- [x] New tests cover all four `draws` formats
- [x] End-to-end tests confirm plots are produced from draws input

### DIFF
--- a/R/helpers-ppc.R
+++ b/R/helpers-ppc.R
@@ -58,15 +58,20 @@ validate_y <- function(y) {
 #' Validate predictions (`yrep` or `ypred`)
 #'
 #' Checks that `predictions` is a numeric matrix, doesn't have any NAs, and has
-#' the correct number of columns.
+#' the correct number of columns. If `predictions` is a `posterior::draws`
+#' object it is first coerced to a matrix.
 #'
-#' @param predictions The user's `yrep` or `ypred` object (SxN matrix).
+#' @param predictions The user's `yrep` or `ypred` object (SxN matrix or a
+#'   `posterior::draws` object).
 #' @param `n_obs` The number of observations (columns) that `predictions` should
 #'   have, if applicable.
 #' @return Either throws an error or returns a numeric matrix.
 #' @noRd
 validate_predictions <- function(predictions, n_obs = NULL) {
-  # sanity checks
+  if (posterior::is_draws(predictions)) {
+    predictions <- posterior::as_draws_matrix(predictions)
+    predictions <- unclass(predictions)
+  }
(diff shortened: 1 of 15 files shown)


jgabry:
Thanks! Yes we should allow posterior::draws objects. I'll try to review this later, I'm a bit short on time right this moment.

jgabry (changes requested):
I ended up having a moment to look at this now. Thanks for this. Just one change to make: I think the documentation for the `yrep` argument should be updated in `man-roxygen/args-y-yrep.R`. 

I made some updates to the tests to make them more directly test what we care about. Otherwise this looks good aside from updating the doc.

jgabry (approved):
Looks good, thanks!

PR #554, opened 2026-07-17
`mcmc_trace`: officially support `highlight` and `alpha` arguments

### LINKED ISSUE #552: mcmc_trace allows highlight but gives a warning
this is a very minor issue: mcmc_trace allows user to highlight one chain but throws a warning.

bayesplot:::.mcmc_trace allows user to highlight one chain
mcmc_trace does likewise, throws a warning but produces the desired output anyway
?mcmc_trace does not mention the highlight option
mcmc_trace_highlight shows points but not lines

jgabry:
closes #552

The original intent was for highlighting to be done only with `mcmc_trace_highlight()`, but since it actually works for `mcmc_trace()` too we should officially support it. Right now we throw a warning saying the arguments are ignored but they actually aren't.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -65,9 +65,9 @@ Suggests:
     survival,
     testthat (>= 3.0.0),
     vdiffr (>= 1.0.2)
-RoxygenNote: 7.3.3
 VignetteBuilder: knitr
 Encoding: UTF-8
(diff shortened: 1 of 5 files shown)


jgabry:
This is pretty straightforward, so going ahead and merging

PR #557, opened 2026-08-24
Prepare for v1.16.0

jgabry:


_pkgdown.yml
jgabry, line 1: @florence-bockting I moved both new vignettes to an "Additional topics" section on the vignettes page on the website. I think that's a good place where we can add more vignettes on specific topics in the future. If we end up having a lot of them we can break them down into more categories. What do you think?

florence-bockting: I like this idea.

PR #560, opened 2026-09-21
Add option to display observed data as points in ppc_km_overlay() and ppc_km_overlay_grouped()

### LINKED ISSUE #558: Option to visualize event times and censoring times as dots in ppc_km_overlay()
I was presenting my work related to predictive model checking for survival models ([Predictive Assessment and Comparison of Bayesian Survival Models for Cancer Recurrence](https://arxiv.org/abs/2601.01662)) at StanCon and somebody asked me about a case where there are so few event times that the Kaplan-Meier curve is far from continuous. In this case, plotting the Kaplan-Meier curve for the observations is clumsy. A similar issue can arise if the time is discretized and there are only few different time points where events are happening, even though the number of events is high.

For these kinds of cases, I think it would be useful if `ppc_km_overlay() `had a binary parameter `dots` that could be used to plot the event times and censoring times as dots instead of plotting the Kaplan-Meier curve. The default value could still be `dots = FALSE`, which plots the Kaplan-Meier curve as before. There are some examples below.

When plotting the Kaplan-Meier curves for these two models, it can be a little bit unintuitive to say which one is fitting and which one is not, since you have to look at the points where the Kaplan-Meier curves fall.

Image
Image

In my opinion, the dots would be a clearer visualization, because from these you can instantly tell, which model fits and which does not.

Image
Image

Another nice thing about the dots is that they can indicate the exact locations where censoring is happening.

Image
Image

I have already implemented this in my own fork so I can make the pull request relatively easily if this is deemed to be a wanted feature for the package.

Sakuski:
Fixes #558 

### Declaration of AI Usage

AI-based tools were used to help to create some parts of the code. All AI modifications were carefully reviewed manually.


jgabry (approved):
Looks good, thanks for the PR! 

R cmd check failed only because it needed the .Rd file regenerated after adding the new argument. I made a couple of very small edits and will merge this when CI passes. Thanks again.

jgabry:
@Sakuski I forgot to ask you to do it as part of this PR, but if you want can you make a new PR adding yourself as a contributor in the DESCRIPTION file? I think this is your third helpful PR, so happy to add you if you want.

